PR #1, opened 2019-02-06
Fix .bazelrc to be compatible with Bazel 0.23.0

philwo:
The two flags --incompatible_remove_native_http_archive and --incompatible_remove_native_git_repository have been enabled by default in a recent release and were subsequently removed, so specifying them in Bazel 0.23.0 will result in an error due to the flags being unknown.

We can just remove them from the .bazelrc now.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,7 +19,3 @@ startup --host_jvm_args=-XX:+UnlockDiagnosticVMOptions --host_jvm_args=-XX:-Inli
 # hits between CLI and IntelliJ builds, and will also be useful if we switch to
 # a shared cache.
 build --experimental_strict_action_env
-
-# This enforces use of the newer rules, as the older native rules will be
-# disabled in a future version of Bazel.
-build --incompatible_remove_native_http_archive --incompatible_remove_native_git_repository


philwo:
FYI @ensonic

ensonic:
> recent release
Do you remeber which one, so that we can tell people to use e.g. >=0.20.0

philwo:
Yes, according to our issue tracker and release notes, the flags were flipped to default in 0.20.0:

https://github.com/bazelbuild/bazel/issues/6569
https://github.com/bazelbuild/bazel/issues/6570

However, AFAIK it shouldn't really matter - `load()`'ing the new http_* rules doesn't depend on manually flipping these flags. The `--incompatible_*` flags are just intended to identify breakages in advance on CI and "should" never be deployed to production or routinely used by developers (although I'm not sure if that's a soft or hard rule, @dslomov should be able to clarify). :)

By the way, not sure if it's interesting for you, but https://github.com/philwo/bazelisk is a helpful wrapper for Bazel, if you often have people building your project who might not have (the correct version of) Bazel installed. You can check it into your repo together with a `.bazelversion` file and it will automatically download and run the matching official release of Bazel for the user's platform.

dslomov:
We do not recommend running production builds with `--incompatible_*` or `--experimental_*` flags (which implies not putting them into checked-in `.bazelrc`s). We can and will remove these flags at random times. 
Please check out our [full breaking change policy](https://docs.bazel.build/versions/master/backward-compatibility.html)

PR #12, opened 2019-05-27
Usability of connecting-robot.md for installation

ct2034:
This makes the commands directly executable


ensonic:
I think this is a good idea, but we should do this consistently for all the docs. I'll check with the other maintainers.

drigz:
I'd be happy to merge this and do the others in a followup.

ensonic:
> I'd be happy to merge this and do the others in a followup.
Works for me. Thanks!

PR #16, opened 2019-09-25
Push prometheus-operator to version 6.11.0

oliver-goetz:
Pushed to most recent version of prometheus-operator because of this issue in grafana https://github.com/grafana/grafana/issues/15739


src/app_charts/prometheus/prometheus-cloud.values.yaml
```diff
@@ -56,6 +58,9 @@ coreDns:
 kubelet:
   enabled: false
 
+# Throws a invalid namespace "kube-system" error during deployment
```
ensonic: typo 'hrows an invalid ', also in prometheus-robot.values.yaml

ensonic:
Thanks. Looks like the 'tlsProxy' is a new feature, I am bit concerned about the kube-proxy one. I do have 'kube-system' namespaces. If you could retry without this change and put the errors on a gist, tha'd be awesome. Thanks!

oliver-goetz:
> Thanks. Looks like the 'tlsProxy' is a new feature (added in version 6.3.0, see [helm/charts@b469d49](https://github.com/helm/charts/commit/b469d491e2cefcfa43b7b2c7c98ebd4259890aa6))
> 
> I am bit concerned about the kube-proxy one. I do have 'kube-system' namespaces. If you could retry without this change and put the errors on a gist, tha'd be awesome. Thanks!

This is the error message from chartassignments CR when not disabling kubeProxy:
`invalid namespace “kube-system” on “/v1/Service/kube-system/kube-kube-proxy”`
The kubeProxy feature is a new one too added in [version 5.17](https://github.com/helm/charts/commit/1c0d7ce99d09471882b69f8863730773d02ee39a#diff-ae0db05747f536183816893bcaf18efb)

src/app_charts/prometheus/prometheus-cloud.values.yaml
oliver-goetz: Can't see it ;)

oliver-goetz: ok now I got it :)

ensonic:
Great, then we've have also not been using this before. If you could fix the typo and mention that both disabled features are only introduced in newer versions in the commit message, I'll merge it.

PR #40, opened 2020-04-01
oauth2-proxy: set --cookie-expire flag to 1h

oliver-goetz:
Bearer tokens expire after 1h, but default cookie expiration of oauth2-proxy is 168h.
This leads to expired Bearer tokens in _oauth2_proxy cookie which makes token-vendor verifying every HTTP request at IAM API when the cookie is older than 1h.
New cookie expiration is in sync with Bearer token lifetime and avoids expired tokens.


drigz:
Thanks, Oliver! I'll pull this in for internal review.

ensonic:
Thanks!
The default lifetime for this call
https://github.com/googlecloudrobotics/core/blob/master/src/java/com/cloudrobotics/tokenvendor/TokenSource.java#L66
is 1 hour
https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken

We should probably make this explicit with a comment in a followup.

PR #41, opened 2020-04-21
Add validation to Robot CRD for properties of status quo

oliver-goetz:
Added `Status` attribute to Go `Robot` CRD data type to improve the reusability of the package
These are the changes of this pull request:
- Use a fixed branch for kubernetes/api, apimachinery and code-generator in generator.sh (kubernetes-1.15 according to the version in bazel WORKSPACE) because there are breaking changes kubernetes-1.18 (New parameter for context in Get, List,... functions).
- Updated Go `RobotStatusRobot `data type to latest version
- Specified status properties in `robots.registry.cloudrobotics.com` to ensure that Robot CRD watcher of approllout controller always receives valid CRs
- Regenerated Go kubernetes clients and deepcopy functions


src/app_charts/base/cloud/registry-crd.yaml
```diff
@@ -54,3 +54,32 @@ spec:
...
+                  enum:
+                    - UNDEFINED
+                    - UNAVAILABLE
+                    - AVAILABLE
+                    - EMERGENCY_STOP
+                    - ERROR
+                lastStateChangeTime:
+                  type: string
+                batteryPercentage:
+                  type: number
```
ensonic, line 78: I would prefer to not have batteryPercentage here. If one wants to do planning based on robot conditions, we'll probably need lots of predicates and this might be too much overload for the status api.

src/app_charts/base/cloud/registry-crd.yaml
```diff
@@ -54,3 +54,32 @@ spec:
...
             project:
               type: string
+        status:
+          type: object
+          properties:
+            cloud:
+              type: object
+            robot:
+              type: object
+              properties:
```
ensonic, line 64: why robot/cloud split here?

src/app_charts/base/cloud/registry-crd.yaml
```diff
@@ -54,3 +54,32 @@ spec:
...
+                  type: string
+                batteryPercentage:
+                  type: number
+                emergencyStopButtonPressed:
+                  type: boolean
+            configuration:
+              type: object
+              properties:
+                trolleyAttached:
+                  type: boolean
```
ensonic, line 85: This is getting very use-case specific. I think we need a different way of storing facts with the robot.

src/go/generate.sh
```diff
@@ -29,8 +35,16 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-export GO111MODULE=off
-go get k8s.io/code-generator/...
+# from "release-1.15" onwards k8s repositories are using go modules
+export GO111MODULE=auto
+rm -Rf "${DIR}/../.gopath/src/k8s.io"
+mkdir -p "${DIR}/../.gopath/src/k8s.io"
+cd "${DIR}/../.gopath/src/k8s.io"
+git clone --branch $K8S_BRANCH $GIT_API
+git clone --branch $K8S_BRANCH $GIT_APIMACHINERY
+git clone --branch $K8S_BRANCH $GIT_CODEGENERATOR
```
ensonic, line 45: can you use
go get k8s.io/code-generator/...@release-1.15
see
https://github.com/kubernetes/client-go/blob/master/INSTALL.md#enabling-go-modules

src/go/pkg/apis/apps/v1alpha1/types.go
```diff
@@ -86,17 +85,17 @@ type ResourceSetPhase string
 
 const (
 	ResourceSetPhasePending ResourceSetPhase = "Pending"
-	ResourceSetPhaseFailed                   = "Failed"
-	ResourceSetPhaseSettled                  = "Settled"
+	ResourceSetPhaseFailed  ResourceSetPhase = "Failed"
+	ResourceSetPhaseSettled ResourceSetPhase = "Settled"
```
ensonic, line 89: this should not be needed

src/go/generate.sh
oliver-goetz: Great, that looks much nicer.
I changed it.

src/app_charts/base/cloud/registry-crd.yaml
oliver-goetz: This is just reflecting the structure from annotations https://github.com/googlecloudrobotics/core/blob/dddd7e00345351e7906c43ac86a08b41adecd880/src/app_charts/base/cloud/registry-crd.yaml#L34-L36
Afaik originally it was used to be able to send missions from cloud cluster to the robot clusters.

src/app_charts/base/cloud/registry-crd.yaml
oliver-goetz: That's fine with me. We probably should remove it from Go and protobuf data type too. What do you think?

src/app_charts/base/cloud/registry-crd.yaml
oliver-goetz: same as comment before

src/app_charts/base/cloud/registry-crd.yaml
ensonic: Ahh, right you're just adding the validation for the existing use.

ensonic:
Alright, could you please reword the title. Instead of ' improve the reusability' maybe say that we add validation to the schema. For the batteryPercentage and trolleAtached we'll keep this in, but please file a new bug, we should think of a more flexible way to transport such data without beeing application specific.

oliver-goetz:
okay done

PR #42, opened 2020-04-22
Avoid Reconcile loops in AppRollout and ChartAssignment controllers because of status updates in Reconciler

oliver-goetz:
Compare `spec` of AppRollout and ChartAssigment CRs in corresponding controllers to identify update events.
`Reconcile` methods update the `status` of their own CRs which could lead to a Reconcile loop if any CR change triggers an update event.

### DIFF
--- a/src/go/pkg/controller/approllout/controller.go
+++ b/src/go/pkg/controller/approllout/controller.go
@@ -80,7 +80,36 @@ func Add(mgr manager.Manager, baseValues chartutil.Values) error {
 
 	err = c.Watch(
 		&source.Kind{Type: &apps.AppRollout{}},
-		&handler.EnqueueRequestForObject{},
+		&handler.Funcs{
+			CreateFunc: func(e event.CreateEvent, q workqueue.RateLimitingInterface) {
+				log.Printf("AppRollout controller received create event for AppRollout %q", e.Meta.GetName())
+				q.Add(reconcile.Request{
+					NamespacedName: types.NamespacedName{Name: e.Meta.GetName()},
+				})
+			},
+			UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+				// Compare spec of old and new AppRollout to avoid Reconcile loop because of status updates of AppRollout
+				change := !reflect.DeepEqual(e.ObjectNew.(*apps.AppRollout).Spec, e.ObjectOld.(*apps.AppRollout).Spec)
+				change = change || e.MetaOld.GetName() != e.MetaNew.GetName()
+				if change {
+					log.Printf("AppRollout controller received update event for AppRollout %q", e.MetaNew.GetName())
+					q.Add(reconcile.Request{
+						NamespacedName: types.NamespacedName{Name: e.MetaNew.GetName()},
+					})
+					if e.MetaOld.GetName() != e.MetaNew.GetName() {
+						q.Add(reconcile.Request{
+							NamespacedName: types.NamespacedName{Name: e.MetaOld.GetName()},
+						})
+					}
+				}
+			},
+			DeleteFunc: func(e event.DeleteEvent, q workqueue.RateLimitingInterface) {
+				log.Printf("AppRollout controller received delete event for AppRollout %q", e.Meta.GetName())
+				q.Add(reconcile.Request{
+					NamespacedName: types.NamespacedName{Name: e.Meta.GetName()},
+				})
+			},
+		},
 	)
 	if err != nil {
 		return errors.Wrap(err, "watch AppRollouts")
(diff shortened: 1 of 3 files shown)


ensonic:
I think this makes sense. We probably want to remove some of the logging to not have too noisy logs.

oliver-goetz:
That's true.

There are many log entries from controller-runtime atm. Thus it would make sense to harmonize the logging. Controller-runtime is using **logr** https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md

Additionally it would be nice to see the different log levels is Stackdriver too. Currently every message is written to stderr, so you see it as an error in there.
Writing info messages to stdout and error messages to stderr or using JSON logging and setting the severity flag could be options for this.

ensonic:
Could you give an example where the loop happened? Also wrt. to the log printfs, i'd probably just drop them for now, but we can discuss alternative log packages separately.

oliver-goetz:
The loop is running all the time. In the current implementation it is hard to identify, because you don't see how the change event was triggered.
Thus, I changed my implementation for testing that it does not filter the change event of AppRollout but write a log entry what was changed. It should work like the current implementation + logging
[cloud-master-5767d8d94c-2vq4t.log](https://github.com/googlecloudrobotics/core/files/4527994/cloud-master-5767d8d94c-2vq4t.log)
. Like this:
```Go
UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
	// Compare spec of old and new AppRollout to avoid Reconcile loop because of status updates of AppRollout
	specChange := !reflect.DeepEqual(e.ObjectNew.(*apps.AppRollout).Spec, e.ObjectOld.(*apps.AppRollout).Spec)
	statusChange := !reflect.DeepEqual(e.ObjectNew.(*apps.AppRollout).Status, e.ObjectOld.(*apps.AppRollout).Status)
	nameChange := e.MetaOld.GetName() != e.MetaNew.GetName()
	log.Printf("AppRollout controller received update event for AppRollout %q", e.MetaNew.GetName())
	log.Printf("AppRollout - SPEC change: %v, STATUS change: %v, NAME change: %v", specChange, statusChange, nameChange)
	q.Add(reconcile.Request{
		NamespacedName: types.NamespacedName{Name: e.MetaNew.GetName()},
	})
	if e.MetaOld.GetName() != e.MetaNew.GetName() {
		q.Add(reconcile.Request{
			NamespacedName: types.NamespacedName{Name: e.MetaOld.GetName()},
		})
	}
},
```
I updated the cluster and let it run for about 10 minutes doing nothing else. You see in the attached log file that there are many update events for AppRollout now where only the status of the CR changed.

oliver-goetz:
A different approach would be the redesign of both `reconcile` methods in a way that the CR status is only updated, if reconciler really changed something or identified a real status change. At the moment most of the time only `lastUpdateTime` changes.
That's probably the way how it is supposed to be and it would make sense to use `handler.EnqueueRequestForObject` in this case.
When I look at `lastUpdateTime` of a deployment which can be weeks in the past, I have the impression this controller is build like that.

ensonic:
I think its worth too go that route. I think we only want a LastUpdateTime if something changed. The semantics would be:
LastTransitionTime is updated only if Status changed
LastUpdateTime is updates is updates if Status or Message changed

Instead of adding more 'if's to [setCondition](https://github.com/googlecloudrobotics/core/blob/master/src/go/pkg/controller/chartassignment/controller.go#L450) I would probably check the 4 call sites, why the call it for a potential no-op.

Sorry I don't understand the link to https://godoc.org/sigs.k8s.io/controller-runtime/pkg/handler and EnqueueRequestForObject()

oliver-goetz:
On these two positions setCondition is called on every reconcile even when nothing changed: https://github.com/googlecloudrobotics/core/blob/ae3f36270fe4464cf6130a06a0e33dfcc0273e58/src/go/pkg/controller/approllout/controller.go#L340-L341 https://github.com/googlecloudrobotics/core/blob/ae3f36270fe4464cf6130a06a0e33dfcc0273e58/src/go/pkg/controller/approllout/controller.go#L346-L347

At those points setCondition only should be called when `got == want` and `condition.Status != core.ConditionTrue`. There are multiple conditions so the last check could look more complicated than adding another `if` to setCondition

Regarding EnqueueRequestForObject I wanted to say that we would not need to filter the change events if we avoid the unnecessary status updates. In this case we could continue using the EnqueueRequestForObject handler for the AppRollout and ChartAssignment CRs.

ensonic:
I agree. an extra check in setCondition might be easier, that checking in setStatus if the condition needs updates.

ensonic:
Thanks. Does it help? Do we still need the change to &handler.Funcs{...} or can we revert that back to &handler.EnqueueRequestForObject{},?

oliver-goetz:
Yes it is working. LastUpdateTime only changes if there are real status changes.
I'll test it with handler.EnqueueRequestForObject now. If it still works, I'll revert the handler changes.

oliver-goetz:
Using handler.EnqueueRequestForObject again Reconcile still runs in a much lower frequency than before.

ensonic:
Awesome. With that it sounds like it is relative harmless to keep the logging in there?

PR #55, opened 2020-09-25
Use Kubernetes 1.16 minimum

oliver-goetz:
Beginning October 6, 2020 GKE starts to migrate Kubernetes control planes to v1.16 automatically.
It seems to be time for an upgrade.

Current cert-manager v0.8.1 is using the deprecated apps/v1beta1 API, which was removed. Thus, I bumped it to v0.10.1 which is the last version supporting certmanager.k8s.io/v1alpha1 API. For later versions we need to migrate to cert-manager.io/v1alpha2. Please see [here](https://cert-manager.io/docs/installation/upgrading/upgrading-0.10-0.11/).

These are the changes of this pull request:
- Kubernetes minimum version raised to 1.16
- Migration from cert-manager v0.8.1 to v0.10.1
- Wait for apiservice `v1beta1.metrics.k8s.io` to become available because otherwise synk fails. This happened every time when terraform was performing the control plane update
- Increase the timeout of `kubectl wait` to 600s. Especially the cert-manager-webhook deployment can take longer than the default timeout when upgrading. In this case deploy.sh failed before.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -217,6 +217,23 @@ function cleanup_old_cert_manager {
       issuers.certmanager.k8s.io \
       clusterissuers.certmanager.k8s.io
   fi
+
+  if [[ "$installed_ver" == 0.8.* ]]; then
+    echo "need to cleanup old version"
+    # see https://cert-manager.io/docs/installation/upgrading/upgrading-0.8-0.9/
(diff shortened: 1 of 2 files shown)


drigz:
Thanks, Oliver! I've pulled this to our internal Gerrit review to see if it affects our integration tests.

PR #58, opened 2020-11-11
Fix invalid escape sequences

Wyverald:
`\(`, `\)`, and `\;` are invalid escape sequences in Starlark. This used to be silently ignored but now throws an error (see https://github.com/bazelbuild/bazel/commit/73402fa4aa5b9de46c9a4042b75e6fb332ad4a7f).


faximan:
Thanks for your contribution!

PR #62, opened 2021-01-13
Add CR_SYNCER_RBAC flag for reducing robot permissions

drigz:
This allows us to reduce the IAM permissions granted to robot-service@, preventing a compromised robot from running workloads in the GKE cluster or on other robots.

Users will need to create RBAC ClusterRoles for their CRs based on the example, add CR_SYNCER_RBAC=true to config.sh, and redeploy their projects. If they fail to configure appropriate policies, they will can check the auth errors in the cr-syncer logs to debug. Once all projects have migrated, we can remove the legacy CR_SYNCER_RBAC=false support.

### DIFF
--- a/config.sh.tmpl
+++ b/config.sh.tmpl
@@ -49,3 +49,9 @@ GCP_ZONE=europe-west1-c
 # Enable google cloud robotics layer 2
 APP_MANAGEMENT=true
 
+# Require explicit RBAC policies for the cr-syncer. If false, the
+# robot-service@ service account has broad permissions in the GKE cluster.
+# If true, you will have to create RBAC policies when enabling the cr-syncer
+# for your own CustomResourceDefinitions. As this is more secure, we plan
+# to enable everywhere after migration.
+CR_SYNCER_RBAC=false
--- a/deploy.sh
+++ b/deploy.sh
@@ -157,6 +157,7 @@ region = "${GCP_REGION}"
 shared_owner_group = "${CLOUD_ROBOTICS_SHARED_OWNER_GROUP}"
 robot_image_reference = "${SOURCE_CONTAINER_REGISTRY}/setup-robot@${ROBOT_IMAGE_DIGEST}"
 crc_version = "${CRC_VERSION}"
+cr_syncer_rbac = "${CR_SYNCER_RBAC}"
 EOF
 
   if [[ -n "${PRIVATE_DOCKER_PROJECTS:-}" ]]; then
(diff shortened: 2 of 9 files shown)


oliver-goetz:
Hi @drigz 
I like the idea of restricting permissions of robot service account. There is a point where this could be extended.

After implementing this, robots would still be able to change the spec of robot CRs. As these include labels which control app rollout, a malicious robot would still be able to change the apps of their own and other robots.
Thus, how about introducing a status sub-resource for robot CRD at this point and restrict update permissions of the robot to it. This would prevent the robots from changing app rollouts.

Probably less relevant is the RobotType CRD. I'm just wondering why robots need update access.

drigz:
Thanks for the suggestion: I'd considered it and decided against it, since I thought that it would require deleting and recreating the CRD. However, your comment made me check that and I realized I was wrong! So I will make the change to the robot CR.

It also occurs to me that the cr-syncer currently wants to update the finalizers, so I should test that this change doesn't cause issues there. Maybe this is the prompt I've been waiting for to remove the finalizer approach as discussed on the other issue.

drigz:
@oliver-goetz Could you check that your robot status controller is happy with a status subresource? You can enable it with `kubectl edit crd robots.registry.cloudrobotics.com` in the cloud & robot cluster, adding `subresources:\nstatus:{}` adjacent to `schema`:

```
spec:
  # SNIP
  versions:
  - name: v1alpha1
    subresources:
      status: {}
    schema:
      # SNIP
```

This comment from the cr-syncer makes me think it might affect how the client needs to update the status:

```
	// We need to make a dedicated UpdateStatus call if the status is defined
	// as an explicit subresource of the CRD.
```

Also, it turns out that the finalizer design is not compatible with this change, so I'll take a look into removing finalizers. Frustratingly, we'll also have to remove the `remote-resource-version` debug feature.

oliver-goetz:
At Kubernetes API Server enabling a status subresource, adds an own endpoint for status update.

The robot CR endpoint is looking like this:
`<host>/apis/registry.cloudrobotics.com/v1alpha1/namespaces/default/robots`

After enabling status subresource there is an additional endpoint: `<host>/apis/registry.cloudrobotics.com/v1alpha1/namespaces/default/robots/status`

From now on the first endpoint ignores the status of the request and the seconds one ignores everything but the status. That's probably the reason why cr-syncer is not able to set finalizers in the cloud cluster anymore.

Our apps are working with and without status subresource. We activated subresources for some of our CRD a while ago and the apps writing robot status identify the right update method from CRD.

Sounds like quite a bunch of changes. Chartassignments (and some of our CRDs) are using status subresources too. Thus, it is at least not because of one single object.
After migration we should not forget to delete the cr-syncer finalizers. Otherwise, existing CRs would remain as orphans forever.

Issue #47, opened 2020-06-08
Offline robot clusters block deployment of recreated approllouts


oliver-goetz:
Offline robot clusters can block approllouts recreated with the same name.
These are the steps to reproduce the error:

1. Setup a scenario with multiple robot clusters
2. Create an approllout which affects all clusters and wait until it is successfully deployed on all robots
3. Turn off one robot cluster (or cut its connection to cloud cluster)
4. Delete the approllout
5. Create the same approllout under the same name again

The freshly recreated approllout CR never gets a status and no new chartassignment CRs are recreated. However when you check the chartassignments CRs the old chartassignment of the offline robot is still present. This happens because the finalizer set by cr-syncer is still present and prevents deletion. This seems to block the creation of new chartassignments for the same approllout.

A workaround is removing the finalizer from orphaned chartassigment CR manually.

drigz:
First thoughts: probably the easiest fix would be to allow the approllout controller to patch existing chartassignments when handling a new approllout. However, that might be inconvenient if you're intentionally using delete-recreate to trigger recreation of the chartassignments to work around a bug somewhere.

It would be nice to fix the "deleting CR that's been synced to a turned-off robot" case, as it's generally quite annoying, although significantly more work. I can imagine replacing the current finalizer-based approach with a GC-based approach, where synced CRs have references to the UIDs of the original CR.

oliver-goetz:
Your second idea sounds very good, because the "deleting CR that's been synced to a turned-off robot" case appears to be the root cause and generates more effort for fixing.
Patching the CRs could cause other issues. Sometimes I delete and recreate an approllout to ensure that a pod is running with the latest version of a docker image. In the Patching cause I could not be sure about this without checking the chartassigments explicitly, because I might recreate the approllout too fast before the chartassigment was deleted.

PR #63, opened 2021-01-19
Replace cr-syncer finalizers with async logic

drigz:
This lets you delete a ChartAssignment from the cloud (upstream) cluster
even when the robot (downstream) cluster is offline and cannot process
the deletion. The downstream resource will be deleted when the robot
comes online.

It also lets you apply RBAC policy to prevent the cr-syncer from
updating specs in the cloud, as it now only needs to update the status.

This will remove finalizers from actively-synced resources, although if
a robot has been taken offline, you will need to manually remove
finalizers and delete its ChartAssignments etc.

### DIFF
--- a/docs/concepts/federation.md
+++ b/docs/concepts/federation.md
@@ -18,18 +18,21 @@ author.
 For federating resources, this means that `spec` and `status` of a resource are owned by at most
 one cluster respectively (possibly the same one). The cluster owning the spec is also the main
 owner of the resource overall and controls its lifecycle, i.e. deletion.
-A resource’s spec is always synced from cluster A to B and its status synced back from B to A.
+A resource’s spec is always synced from an upstream cluster to a downstream
+cluster and its status synced back from downstream to upstream.
 
 All resources of a specific type may either be synchronized to all robots or to exactly one robot.
 There is no direct synchronization between robots. However, a robot may create a resource in the
 cloud cluster that will be distributed to other robots.
 
-If a resource owned by cluster A has been synchronized to one or more clusters, it can only be
-permanently deleted in cluster A. It’s full deletion in A will be blocked until it has been
-propagated and completed across all clusters that held a copy of the resource.
+If a resource owned by the upstream cluster has been synchronized to one or more
+downstream clusters, it can only be permanently deleted upstream: if deleted
+downstream, it will be recreated. If deleted in the upstream cluster, it will be
+asynchronously deleted in other clusters that hold a copy of the resource.
+Upstream deletion can complete before the downstream resource is deleted.
 
 ## cr-syncer
-The cr-syncer component (Custom Resource Syncer) is a controller that runs once inside each robot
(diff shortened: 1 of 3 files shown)


drigz:
@oliver-goetz I think this would address #47 (as well as unblocking the RBAC change). Could you take a look at the new doc and see if it makes sense to you?

oliver-goetz:
@drigz your documentation changes sound reasonable 👍  
Maybe a hint for the upgrade process would be helpful, that you should update all robots with `CR_SYNCER_RBAC=false` before you enable it. Otherwise cr-syncer would not be able to remove all finalizers. I think the explanation of the setting in `config.sh.tmpl` could be the right place for such a comment.

I tested the changes too, still with `CR_SYNCER_RBAC=false`. cr-syncer did what it should do and removed the finalizers. There are two things I observed:

cr-syncer [deletes all its own finalizers](https://github.com/googlecloudrobotics/core/pull/63/files#diff-8393e2ec32d98a27fafaa9b5555315a8b2460ec69a124bda5774646132a11a8dR101) independent from the robot name. This could lead to an infinite update loop for CRs which are synced to multiple robots like RobotType as long as some robots are updated and others not. Updated robots remove all finalizers of these CRs and not yet upgraded robots add their own finalizer...and again...

There are some, not very frequent error messages where I don't know if they are related to these changes or not.
```
2021/01/23 14:44:26 Got add event from upstream for OrderAuction 1710.1611413065-mir-robot01@v182225686
2021/01/23 14:44:26 Got add event from downstream for OrderAuction 1710.1611413065-mir-robot01@v4923773
2021/01/23 14:44:26 Got update event from downstream for OrderAuction 1710.1611413065-mir-robot01@v4923776
2021/01/23 14:44:26 Copied OrderAuction 1710.1611413065-mir-robot01 status@v4923773 to upstream@v182225694
2021/01/23 14:44:26 Got update event from upstream for OrderAuction 1710.1611413065-mir-robot01@v182225694
2021/01/23 14:44:26 Syncing key "default/1710.1611413065-mir-robot01" from queue "downstream" failed: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)
2021/01/23 14:44:27 Copied OrderAuction 1710.1611413065-mir-robot01 status@v4923776 to upstream@v182225698
2021/01/23 14:44:27 Got update event from upstream for OrderAuction 1710.1611413065-mir-robot01@v182225698
``` 
The CR was synced correctly in the end and I just noticed the error because I was checking the logs during upgrade.

drigz:
Great comments, thank you! I hadn't thought of the issue with staggered updates, it would be cleaner to only delete the finalizer for the given robot. Any finalizers for offline robots can easily be manually removed. I think I was able to fix the "panic" error, I suspect that's caused by a "remote resource has been updated, please retry" issue, and I made a bad refactor that caused the nil dereference.

I'll do some manual testing tomorrow and hopefully submit both this and the CR_SYNCER_RBAC change.

Issue #67, opened 2021-06-25
I/O Timeouts to Kubernetes API in Cloud Cluster lead to cr-syncer crash loop


oliver-goetz:
I created a little test case to check how cr-syncer reacts on connectivity issues

- There is a robot cluster deployed in a VM in GCE
- On this robot cluster the status of the robot CR is updated every 2 seconds
- I created a firewall rule to block access from this VM to the Cloud Robotics ingress IP
- The firewall rule only applies on new connections, so I killed the nginx pod to force cr-syncer to reconnect 

The first three lines of the following log are the regular log entries when connection was available. After restarting nginx there are only downstream events left which sounds reasonable.
After a couple of more events there is a i/o timeout error and cr-syncer restarts without further notice. 

```
2021/06/24 15:54:36 Got update event from downstream for Robot mir-robot01@v9508595
2021/06/24 15:54:36 Got update event from upstream for Robot mir-robot01@v334026459
2021/06/24 15:54:36 Copied Robot mir-robot01 status@v9508595 to upstream@v334026459
2021/06/24 15:54:38 Got update event from downstream for Robot mir-robot01@v9508599
2021/06/24 15:54:40 Got update event from downstream for Robot mir-robot01@v9508602
2021/06/24 15:54:42 Got update event from downstream for Robot mir-robot01@v9508606
2021/06/24 15:54:44 Got update event from downstream for Robot mir-robot01@v9508609
2021/06/24 15:54:46 Got update event from downstream for Robot mir-robot01@v9508612
2021/06/24 15:54:48 Got update event from downstream for Robot mir-robot01@v9508616
2021/06/24 15:54:50 Got update event from downstream for Robot mir-robot01@v9508619
2021/06/24 15:54:52 Got update event from downstream for Robot mir-robot01@v9508624
2021/06/24 15:54:54 Got update event from downstream for Robot mir-robot01@v9508627
2021/06/24 15:54:56 Got update event from downstream for Robot mir-robot01@v9508630
2021/06/24 15:54:58 Got update event from downstream for Robot mir-robot01@v9508634
2021/06/24 15:55:00 Got update event from downstream for Robot mir-robot01@v9508637
2021/06/24 15:55:02 Got update event from downstream for Robot mir-robot01@v9508642
2021/06/24 15:55:04 Got update event from downstream for Robot mir-robot01@v9508645
2021/06/24 15:55:06 Got update event from downstream for Robot mir-robot01@v9508648
[...]

PR #68, opened 2021-06-28
Improve resilience of cr-syncer regarding network errors

oliver-goetz:
@drigz, this is the follow up to issue https://github.com/googlecloudrobotics/core/issues/67

It changes the limiter of the queue which is source for changes in cloud cluster to `ItemFastSlowRateLimiter` that there is no exponential backoff. Exponential backoff could be the reason for long recovery times on unstable network connectivity.
`ItemFastSlowRateLimiter` is configured to do 5 fast retries with 500ms backoff time, followed by a series of slow retries with 5 seconds backoff time.

The limiter of the queue for updating CRs in der robot cluster remains `DefaultControllerRateLimiter` because we don't expect connectivity issues there.

### DIFF
--- a/src/go/cmd/cr-syncer/syncer.go
+++ b/src/go/cmd/cr-syncer/syncer.go
@@ -20,6 +20,7 @@ import (
 	"log"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/pkg/errors"
 	"go.opencensus.io/stats"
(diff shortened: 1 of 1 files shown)


drigz:
Thanks! I think the biggest risk here is that if some unrelated error happens, we could get (number of CRs * number of robots / 5) QPS on the cloud API server. At larger deployments that could cause a problem, but I don't see it causing a problem for our comparatively small number of CRs and robots.

Issue #64, opened 2021-01-26
cr-syncer does not retry on failed updates


oliver-goetz:
We find quite a number of these errors in our cr-syncer logs.
```
2021/01/26 16:40:16 Syncing key "default/1710.2053438" from queue "downstream" failed: WarehouseOrder default/1710.2053438 @ 184236994: update status failed: Operation cannot be fulfilled on warehouseorders.ewm.sap.com "1710.2053438": the object has been modified; please apply your changes to the latest version and try again
```
cr-syncer does not retry applying the change as the error message suggests. If there are other update events from downstream cluster, it will get in sync again. But if there are no more change events from downstream for that CR it remains in a not synced state. 
When a CR is in this state it even does sync changes from upstream cluster anymore.
It remains in this inconsistent state until cr-syncer is restarted.

oliver-goetz:
I made another observation. Sometimes the status of robots is not updated in cloud cluster, even though cr-syncer logs give the impression that the status was synced.
I made a little [screen cap video](https://sap-my.sharepoint.com/:v:/p/o_goetz/EaGA1ev9nIJOoSWJIebn3GoBKa54cfrIKlyHTjNHkLbkSg?e=eaOhqp). The upper left terminal shows cr-syncer logs of mir-robot01, the upper right terminal watches the robot CR of mir-robot01 directly on the robot. Finally the lower terminal watches the CR of mir-robot01 in the cloud cluster.

You see that on the robot its status is continuously updating, cr-syncer receives the update event, but it is not synced to cloud cluster.
Additionally you see some of the _syncing key issues_ mentioned before, but they are probably not related.

Looking into the nginx log of cloud cluster, there is a 409 error like this in the meantime for each robot and then no more log entries for those CRs for a while. Syncing recovers a while, but I don't know the concrete point in time. It is rather a matter of minutes than seconds until it recovers.

```
111.111.111.111 - [111.111.111.111] - - [27/Jan/2021:14:30:42 +0000] "PUT /apis/core.kubernetes/apis/registry.cloudrobotics.com/v1alpha1/namespaces/default/robots/mir-robot01/status HTTP/2.0" 409 374 "-" "cr-syncer.binary/v0.0.0 (linux/amd64) kubernetes/$Format" 968 0.003 [default-kubernetes-443] 222.222.222.222:443 374 0.004 409 61189d593e12555bb2278915a242e3da
```

[...]

PR #69, opened 2021-07-22
Restart cr-syncer informers on too many subsequent conflict errors

oliver-goetz:
Following up to this issue https://github.com/googlecloudrobotics/core/issues/64

When trying to run cloud-robotics with Istio ingresses I realised that is is hard to guarantee that no issues as in #64 happen again as long as client-go does not identify connection drops reliably.
One way to identify that a watcher died are many subsequent conflict errors when cr-syncers updates the remote cluster
```
2021/01/26 16:40:16 Syncing key "default/1710.2053438" from queue "downstream" failed: WarehouseOrder default/1710.2053438 @ 184236994: update status failed: Operation cannot be fulfilled on warehouseorders.ewm.sap.com "1710.2053438": the object has been modified; please apply your changes to the latest version and try again
``` 
This change restarts informers for one CRD in case 5 subsequent conflict errors occur. This is more efficient than restarting the whole cr-syncer because it could happen for single CRDs.

### DIFF
--- a/src/go/cmd/cr-syncer/syncer.go
+++ b/src/go/cmd/cr-syncer/syncer.go
@@ -52,6 +52,8 @@ const (
 	// is disabled, otherwise the status and annotation cannot be updated
 	// in a single request.
 	annotationResourceVersion = "cr-syncer.cloudrobotics.com/remote-resource-version"
+
+	clusterNameCloud = "cloud"
 )
 
 var (
@@ -137,6 +139,9 @@ type crSyncer struct {
 	downstreamInf   cache.SharedIndexInformer
 	upstreamQueue   workqueue.RateLimitingInterface
 	downstreamQueue workqueue.RateLimitingInterface
+	infDone         chan struct{}
+
+	conflictErrors int
 
 	done chan struct{} // Terminates all background processes.
 }
@@ -193,7 +198,7 @@ func newCRSyncer(
 	}
 	switch src := annotations[annotationSpecSource]; src {
 	case "robot":
-		s.clusterName = "cloud"
+		s.clusterName = clusterNameCloud
 		// Swap upstream and downstream if the robot is the spec source.
 		s.upstream, s.downstream = s.downstream, s.upstream
 		// Use DefaultControllerRateLimiter for queue with destination robot and ItemFastSlowRateLimiter for queue with destination cloud to improve resilience regarding network errors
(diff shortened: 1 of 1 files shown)


src/go/cmd/cr-syncer/syncer.go
```diff
@@ -52,6 +52,8 @@ const (
 	// is disabled, otherwise the status and annotation cannot be updated
 	// in a single request.
 	annotationResourceVersion = "cr-syncer.cloudrobotics.com/remote-resource-version"
+
+	clusterNameCloud = "cloud"
```
ensonic, line 56: cloudClusterName

src/go/cmd/cr-syncer/syncer.go
```diff
@@ -295,6 +320,24 @@ func (s *crSyncer) processNextWorkItem(
 	}
 	defer q.Done(key)
 
+	// Restart informers on too many conflict errors
+	// client-go does not reliably recognize when watch calls are closed by remote API server
+	// cr-syncer is able to detect that when updating CRs on remote API server when there are multiple subsequent conflict errors (HTTP 409)
+	// like "...please apply your changes to the latest version and try again"
+	// This could occur at watchers of single CRDs while others keep working. Thus, it is less resource intensive just restarting informers of the affected CRDs rather than whoel cr-syncer
+	// Errors are counted in syncUpstream and syncDownstream functions
+	if s.conflictErrors >= 5 {
```
ensonic, line 329: Please make 5 a constant or ideally a flag, so that it is easy to play with the value by modifying the yaml. Using 5 as the default make sense.

src/go/cmd/cr-syncer/syncer.go
oliver-goetz: done

src/go/cmd/cr-syncer/syncer.go
oliver-goetz: there is a flag called `conflict-error-limit` now

oliver-goetz:
Hi @ensonic, a friendly reminder

drigz:
I'm afraid Stefan is on vacation at the moment, but he'll be back later this week. Let me know if this is urgent and I can try to take over.

oliver-goetz:
thanks for notifying me.
It is not urgent. I'm just back from my vacation an checking what happened in the meantime 😄

drigz:
Ah, in that case, I think the answer is "some walks in the Austrian Alps" but Stefan can confirm :)

oliver-goetz:
that sounds good 😄

PR #71, opened 2021-08-06
Fix Helm download location

fweikert:
The old binary is no longer accessible since Helm moved their releases to a new GCS bucket. I got the URL from https://github.com/helm/helm/releases/tag/v2.17.0


drigz:
Thank you, Florian! I'll pull this into Gerrit and submit it.

PR #79, opened 2021-09-28
Update to cert-manager v1.5.3

oliver-goetz:
This PR updates cert-manager to v1.5.3 and cleans up the the old version 0.10.1

There are new CRDs for cert-manager with the new apiVersion `cert-manager.io/v1`. Thus, old certificates will be removed by this update. Default cloud-robotics certificates and issuers are recreated, others have to be migrated manually.


ensonic:
I think something is missing in the migration. When I apply this to an existing installation, I ran into:
```
Initialized successfully
synk init done
No resources found
checking for old cert manager ..
have cert manager 1.5.3
No resources found
namespace/default unchanged
installing base-cloud to gke_robco-integration-test_europe-west1-c_cloud-robotics...
apply files: set default namespaces: discover server resources: unable to retrieve the complete list of server APIs: webhook.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request
Synk failed for base-cloud
```

oliver-goetz:
> I think something is missing in the migration. When I apply this to an existing installation, I ran into:
> 
> ```
> Initialized successfully
> synk init done
> No resources found
> checking for old cert manager ..
> have cert manager 1.5.3
> No resources found
> namespace/default unchanged
> installing base-cloud to gke_robco-integration-test_europe-west1-c_cloud-robotics...
> apply files: set default namespaces: discover server resources: unable to retrieve the complete list of server APIs: webhook.certmanager.k8s.io/v1beta1: the server is currently unable to handle the request
> Synk failed for base-cloud
> ```

The webhook of the old version seems to block your deployment. It should be deleted by the migration script [here](https://github.com/googlecloudrobotics/core/pull/79/files#diff-30883d1df2fc62f59d66255a70dc05f60d3d4ef4d3957e6e3e38e62f6e471bd0R276)
Is the output from the second time you run the script? I'm wondering about that because of `have cert manager 1.5.3` which is already the new version of cert-manager

ensonic:
Here is the full log of the first run, where it did the upgrade:
https://gist.github.com/ensonic/bd57253f37916e8a14111de11e24d1c6

oliver-goetz:
It seems the webhook was [successfully deleted](https://gist.github.com/ensonic/bd57253f37916e8a14111de11e24d1c6#file-gistfile1-txt-L397-L398) in the first place.
There might be some things left of the old installation so there is a certificate of an [unknown authority](https://gist.github.com/ensonic/bd57253f37916e8a14111de11e24d1c6#file-gistfile1-txt-L502-L506) which makes the deployment fail.
I don't know why the webhook is back afterwards. Maybe it is a kind of "rollback" from synk, but it seems to be the symptom and not the root cause. 
cert-manager cannot recover from the "unknown authority" error. I had this error before but I thought I removed all the certificates. Apparently that's not the case. I'll test it again with a new migration cluster

ensonic:
I could manually fix it by:
```
alias kc="kubectl --context=gke_XXXX_europe-west1-c_cloud-robotics"
```
and then manually rerunning the deletion steps. I wonder if we need to wait a little after the deletion to ensure things are gone before reinstalling?

oliver-goetz:
We could wait until the pods are terminated after [deleting the deployments](https://github.com/googlecloudrobotics/core/pull/79/files#diff-30883d1df2fc62f59d66255a70dc05f60d3d4ef4d3957e6e3e38e62f6e471bd0R255-R258)
It have been the case that the pods recreated the webhook CAs which are deleted [just some lines after](https://github.com/googlecloudrobotics/core/pull/79/files#diff-30883d1df2fc62f59d66255a70dc05f60d3d4ef4d3957e6e3e38e62f6e471bd0R264) before they have been terminated. This could be the reason for the unknown authority error.

ensonic:
Lets try that (adding `--wait=true` to kubectl delete calls). I can retry the migration only personal project.

oliver-goetz:
The migration script now waits until the old cert-manager pods are deleted. I used `kubectl wait` instead of `--wait=true` because the later just waits until the deployment is deleted. The pods need a bit longer to terminate.
`--timeout=35s` was chosen to be a bit longer then default termination grace period of 30s.

ensonic:
Awesome. This time it worked great.

PR #80, opened 2021-09-28
Update cloud-robotics CRDs to v1

oliver-goetz:
[The apiextensions.k8s.io/v1beta1 API version of CustomResourceDefinition is no longer served as of v1.22](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122)

This PR migrates cloud-robotics.com CRDs to `apiextensions.k8s.io/v1`

### DIFF
--- a/docs/how-to/creating-declarative-api.md
+++ b/docs/how-to/creating-declarative-api.md
@@ -234,7 +234,7 @@ Create a file called `charge-crd.yaml` with the following contents:
 
 [embedmd]:# (examples/charge-service/charge-crd.yaml yaml)
 ```yaml
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: chargeactions.example.com
@@ -242,14 +242,17 @@ metadata:
     cr-syncer.cloudrobotics.com/spec-source: cloud
 spec:
   group: example.com
-  version: v1
   names:
     kind: ChargeAction
     plural: chargeactions
     singular: chargeaction
   scope: Namespaced
-  subresources:
-    status: {}
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+      subresources:
+        status: {}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/docs/how-to/examples/charge-service/charge-crd.yaml
+++ b/docs/how-to/examples/charge-service/charge-crd.yaml
@@ -1,4 +1,4 @@
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: chargeactions.example.com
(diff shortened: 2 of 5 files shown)


ensonic:
Awesome, could you please also fix it in this test:
https://github.com/googlecloudrobotics/core/blob/master/src/go/tests/apps/apps_test.go#L157

oliver-goetz:
> Awesome, could you please also fix it in this test: https://github.com/googlecloudrobotics/core/blob/master/src/go/tests/apps/apps_test.go#L157 and some of the other places: https://github.com/googlecloudrobotics/core/search?q=cloudrobotics.com%2Fv1alpha1

what's wrong with those tests? All tests are passing on my machine. The CRs themselves should not be affected.
However, I don't know how to run the integration test. Thus, I might missed an issue.

ensonic:
The integration tests are not run by default if you run `bazel test //...` since they will require a cloud-cluster to deploy things.

We run them against a test-project like this:
```
REGISTRY="gcr.io/<your-project>"
bazel run //src/app_charts:push "${REGISTRY}"
ACCESS_TOKEN="$(gcloud auth application-default print-access-token)"
  
bazel test \
  --test_env ACCESS_TOKEN="${ACCESS_TOKEN}" \
  --test_env REGISTRY="${REGISTRY}" \
  --test_tag_filters="requires-docker" //...
```

oliver-goetz:
There have been some things to change in order to make things work
1. In `apiextensions.k8s.io/v1beta1` `preserveUnknownFields` is by default true whereas in `apiextensions.k8s.io/v1beta1` it is false. I added some `x-kubernetes-preserve-unknown-fields: true` flags where properties of type object are not entirely specified to prevent pruning here. 
2. As a result of the changed setting in 1. it turned out that apps-crd.yaml and registry-crd.yaml were not completely in sync. Thus some properties have been pruned accidentally. I changed the CRD definition from the yaml files that they fit with the go  structures now.
3. The oauth_proxy helm chart which was used for the integration test was [deprecated and removed](https://github.com/k8s-at-home/charts/pull/1132). It is now replaced by its successor.

However, I was still not able to run the integration test because chart-assignment-controller was in a crash loop with this error message.
```
➜  apps git:(update-cloud-robotics-crds) ✗ KUBECONFIG=/tmp/kubeconfig-131896865 kubectl logs chart-assignment-controller-84777bdb6-h2xzg
2021/09/30 15:40:22 Failed to create the Stackdriver exporter: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```
For testing purposes I removed the `--trace-stackdriver-project-id` parameter from chart-assignment-controller deployment. After that the integration test ran successfully.
The last issue seems to be unrelated to the changes of this PR and might depend on the setup of my local machine. Could check that on your end @ensonic 

I committed everything except by temporary deployment change.

oliver-goetz:
Btw. as a result of default value false for `preserveUnknownFields` all custom properties in the Robot/RobotType CRs will be deleted when applying this change.

oliver-goetz:
`ResourceSet` created by synk should be already using `apiextensions.k8s.io/v1` because it is using client-go version 1.17.
However, it does [not have a real schema and is allowing anything](https://github.com/googlecloudrobotics/core/blob/master/src/go/pkg/synk/synk.go#L155-L167).

PR #81, opened 2021-10-04
Fix robot chart in app CRD

oliver-goetz:
@ensonic there was one more little inconsistency between apps-crd.yaml and the go structure I did not find in the first place. It's just one letter, but in the end it removes all the robot charts from App CRs.

### DIFF
--- a/src/app_charts/base/cloud/apps-crd.yaml
+++ b/src/app_charts/base/cloud/apps-crd.yaml
@@ -36,7 +36,7 @@ spec:
                           type: string
                         inline:
                           type: string
-                    robots:
+                    robot:
                       type: object
                       properties:
                         name:


src/app_charts/base/cloud/apps-crd.yaml
```diff
@@ -36,7 +36,7 @@ spec:
                           type: string
                         inline:
                           type: string
-                    robots:
+                    robot:
```
ensonic, line 39: In line 94 below it is also robots:

ensonic: I think robots (plural) is correct and also what we actually use everywhere

oliver-goetz: Prometheus for example calls it [prometheus-robot](https://github.com/googlecloudrobotics/core/blob/master/src/app_charts/prometheus/BUILD.bazel#L40), if I understand the underlying bazel function correctly.
And the [go type](https://github.com/googlecloudrobotics/core/blob/master/src/go/pkg/apis/apps/v1alpha1/types.go#L129) calls it robot too. That's why we call it robot in all our app charts too.

oliver-goetz: We use `robot` in [App CRs](https://github.com/SAP/ewm-cloud-robotics/blob/master/applayer-yaml/ewm-robot-controller/app.yaml#L13) since the beginning and it would be more effort to change it here.
In the end there was no other way to use the name of the Go structure and we just did not notice the inconsistency with App CRD specification from app-crd.yaml because CRDs in v1beta1 preserved unknown fields.

ensonic: RIght, this is for the app! Makes sense.

PR #83, opened 2021-10-09
Update controller-runtime to v0.10.2 / client-go to v0.22.2

oliver-goetz:
This PR updates controller-runtime to the latest version v0.10.2 and client-go, apimachinery etc. to v0.22.2.

There have been lots of contexts to be added and the CRD client libraries had to be regenerated too.

1. build is working
2. test are running successfully. I disabled a [synk test case](https://github.com/googlecloudrobotics/core/blob/d29b27fc5df0b6364d03db6aead3b9e510fac3ed/src/go/pkg/synk/synk_test.go#L367-L375). There are issues when running strategic merge patches against a dynamic fake client. When run the same case against a real cluster, it is working.
3. integration test is not run yet (there is still the issue that I have to remove stackdriver from chart-assignment-controller deployment temporary to make it work on my machine) and might need an update too.
4. I tested it successfully on a cluster with 10 robots with tasks like create, update, delete approllouts and watching cr-syncer logs for half an hour.

I did not test yet if cr-syncer identifies connection issues more reliable after those changes. It might be a good idea to test this in a second step after all these changes are completely tested and included.


oliver-goetz:
Integration Test is passing too 😄

oliver-goetz:
> Regarding the synk patch, I've found this thread: [kubernetes/kubernetes#58017](https://github.com/kubernetes/kubernetes/issues/58017), with some hints (https://github.com/helm/helm/pull/3459/files) that we might need to update synk here:
> 
> https://github.com/googlecloudrobotics/core/blob/d29b27fc5df0b6364d03db6aead3b9e510fac3ed/src/go/pkg/synk/synk.go#L709

I agree we should change this. I'm not aware of any issues here, but the Kubernetes documentation is saying [Note: Strategic merge patch is not supported for custom resources.
](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#summary)

The reason for it might be hidden in [this note](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#notes-on-the-strategic-merge-patch)
I could imagine the reason for this is that unstructured objects do not provide a `patchStrategy`. I'm not sure though.

PR #84, opened 2021-10-15
Add timeout option for cr-syncer

oliver-goetz:
We are still facing some connection issues using istio ingress. The error message suggests that the reason for these issues could be closed connections between ingress and kubernetes api server. Even though the message tells us "NO_ERROR" the connection is closed afterwards.
```
W1013 22:20:07.214926       1 reflector.go:441] go/cmd/cr-syncer/syncer.go:259: watch of *unstructured.Unstructured ended with: an error on the server ("unable to decode an event from the watch stream: stream error: stream ID 3605; NO_ERROR") has prevented the request from succeeding
```

I found a good option to deal with those kinds of issues by introducing a timeout for the watch call.
It is implemented at two places 
1. The options of List & Watch function
2. The remote client

When implementing the timeout just at the watch function. it is included into an URL parameter of the watch call. With this parameter set, the API server is closing the connection, but the timeout is not enforced on client side. Thus, if the connection breaks before the timeout cr-syncer does not recognize it.
By default the timeout url parameter seems to set to a random value between 4 and 10 minutes (estimated).

Timeout is disabled by default, so cr-syncer is not able to recognize an eventual timeout by himself.
When implementing it in the remote client only, there are warnings on each reconnect
```
W1015 15:52:02.629074       1 transport.go:288] Unable to cancel request for *main.ctxRoundTripper
```

That's why I implemented it at both places.
There is a new parameter `--timeout` where you can set the timeout in order to optimize the settings. By default it is set to 300 seconds. This should not stress the api server too much.
I set the timeout in the client to `timeout +5` to give it some time for a graceful closing.

Additionally, I removed the `configureTransport` function which is now the default setting at http client in apimachinery.
Btw. this feature is not detecting the lost connection in our case. Probably the istio ingress is answering the http ping requests even when its connection to the backend is broken.

### DIFF
--- a/src/go/cmd/cr-syncer/main.go
+++ b/src/go/cmd/cr-syncer/main.go
@@ -81,7 +81,7 @@ var (
 	verbose            = flag.Bool("verbose", false, "Enable verbose logging")
 	listenAddr         = flag.String("listen-address", ":80", "HTTP listen address")
 	conflictErrorLimit = flag.Int("conflict-error-limit", 5, "Number of consecutive conflict errors before informer is restarted")
-	timeout            = flag.Int64("timeout", 300, "Timeout for CR watch calls")
+	timeout            = flag.Int64("timeout", 300, "Timeout for CR watch calls in seconds")
 
 	sizeDistribution    = view.Distribution(0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 33554432)
 	latencyDistribution = view.Distribution(0, 1, 2, 5, 10, 15, 25, 50, 100, 200, 400, 800, 1500, 3000, 6000)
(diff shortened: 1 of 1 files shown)


ensonic:
Wrt. to your last comment. Does the change help your case?

oliver-goetz:
Unfortunately the latest version of kubernetes library did not help in our case. The changes introduced in [v0.18.6 and v0.19.4 ](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/net/http.go#L175-L191)are the same @drigz implemented in cr-syncer for the remote connection before.
I just removed it in this PR, because it is duplicated code now.

Apparently the different ingress behave differently in our scenario `cr-syncer -> ingress -> kubernetes api server` on HTTP protocol level. We are facing these issues using istio, but nginx is working fine now. However, in the past we had some [issues with nginx](https://github.com/googlecloudrobotics/core/commit/b2e5d16b08b66bbd910a5df2ac0a74a565b3744d) too.
It might be possible that this is related to something else like different loadbalancers. This is pretty hard to find debug, because I don't find a real error message in ingress.

I'm wondering how the setup would work if ingress forwards on TCP protocol level rather than on HTTP level. In this case there would be still two connection as before, but the HTTP/2 ping check mentioned before could be answered by API server only and not by ingress anymore. 
This would need some changes, because in this case URL rewriting in ingress would not be possible anymore, so we would have to open a dedicated port (or host) for cr-syncer's connection to the API server. TLS termination at ingress should still be possible. 
I will test in our setup if this really makes a difference. What do you think?

However, even then a dedicated timeout for the remote connection sounds reasonable to me.

PR #85, opened 2021-10-26
Add clusterroles and dedicated service accounts for app management apps / remove permissive binding

oliver-goetz:
This PR adds cluster roles and for app-rollout and chart-assignment-controllers and starts them with dedicated service accounts. **Additionally the permissive-binding is removed during deployment.**

The cluster roles for app-rollout-controller are pretty easy and just include additional permissions for `robots`, `apps`, `approllouts` and `chartassignments` CRDs.

Chart-assignment-controller has now permissions for `chartassignments` and `resourcesets` CRDs and for namespaces. For convenience the `cluster-admin` role is assigned to its service account too, because we do not really have a concept yet, which Kubernetes objects we would like to be managed by chart-assignment-controller.

However, this change allows to investigate roles and permissions of other apps now, which was somehow pointless while a permissive-binding was existing. Thus, this PR might lead to "permission denied" in other apps.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -333,13 +333,9 @@ function helm_charts {
     ca_key=$(openssl base64 -A < ${certdir}/ca.key)
   fi
 
-  # Create a permissive policy if none exists yet. This allows code running in the
-  # cluster to administer it.
-  if ! kc get clusterrolebinding permissive-binding &>/dev/null; then
-    kc create clusterrolebinding permissive-binding \
-      --clusterrole=cluster-admin \
-      --user=admin \
-      --group=system:serviceaccounts
+  # Delete permissive binding if it exists because from previous deployments
+  if kc get clusterrolebinding permissive-binding &>/dev/null; then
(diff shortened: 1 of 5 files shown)


oliver-goetz:
I forgot roles & permissions for nginx in the first place 😅 
That's fixed now.

drigz:
Hi Oliver, sorry for taking so long to look at this! It looks good although I'd like to test it on our internal deployments before merging to check we won't break anything.

A side note: have you also reduced the privilege of the [default GCP Service Account](https://cloud.google.com/compute/docs/access/service-accounts#default_service_account)? If you're still running CRC in GKE, by default any pod has Editor access to the GCP project. I'm looking into how to change this, likely the first step will be to replace the Editor binding with some more specific bindings. For the base CRC workloads, I believe roles/cloudiot.provisioner is all that is required, but I haven't tested this yet. If this isn't relevant to your deployment, sorry for the noise.

PR #87, opened 2021-12-21
Check availability of served CRDs only

oliver-goetz:
This tiny patch excludes versions of CRDs which are not served from availability check in `synk`. Unserved version are never available and they currently block deployments by synk.
I was facing this issue when trying to upgrade cert-manager to v1.6.1 because they stop serving CRD versions v1alpha1, v1alpha2 and v1alpha3.


ensonic:
Thanks!

PR #101, opened 2023-02-23
Add support for Bazel RBE

faximan:
Corresponds to insrc #87.

Will be used by CI/CD because GitHub runners are not powerful enough for remote-cache-only builds.


ensonic (approved):
Thanks!

PR #102, opened 2023-02-23
Add CI/CD based on GitHub actions

faximan:
Sorry for the large PR.

This is copying the CI/CD setup from insrc with minimal changes (adapting some paths, using a different cloud project for RBE). I think splitting this up into multiple PRs would be more work than what it is worth, please let me know if you disagree.

I had to copy the `deployments/` folders for robco-integration-test and robco-navtest as they are needed by the CI. However, I removed the Oauth secrets from config.sh, which don't seem needed by the test. My plan for now is to keep the duplicate setup, which I expect to change very rarely.


.github/ci/deployments/common/k8s-relay-rollout.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apps.cloudrobotics.com/v1alpha1
```
ensonic, line 1: I think we can drop all these rollout yaml files (deployments/common and deployments/robco-integration-test/kubernetes/). The would be referenced by the deploy.sh on the infrastructure folder in insrc.

faximan: They are referenced from integration_test.sh part of this PR ("Deploy approllouts and configuration"), but maybe we can drop that and make that part of the apps CI. Would you prefer that?

.github/ci/integration_test.sh
```diff
@@ -0,0 +1,150 @@
...
+  while ! echo "lock" | gsutil -q -h "x-goog-if-generation-match:0" cp - $LOCK_OBJECT
+  do
+    : "lock: failed to obtain lock, retrying in $LOCK_BACKOFF_SECONDS seconds"
+    : "Note to build cop: if you think there is a stale lock, run:"
+    : "    gsutil rm $LOCK_OBJECT"
+    : "This can occur when a previous job timed out or was canceled while"
+    : "holding the lock."
+    sleep $LOCK_BACKOFF_SECONDS
+  done
+  # TODO(rodrigoq): if the build is cancelled by GitHub, the lock is not
```
drigz, line 24: Out of curiosity, does this script run in GitHub Actions or Cloud Build?

drigz: Answered my question by reading further.

deploy.sh
```diff
@@ -429,7 +430,7 @@ function create {
 function delete {
   include_config_and_defaults $1
   if is_source_install; then
-    bazel build "@hashicorp_terraform//:terraform"
+    bazel ${BAZEL_FLAGS} "@hashicorp_terraform//:terraform"
```
drigz, line 433: Missing `build`?

drigz (approved):

deploy.sh
faximan: Wow, nice catch! Done.

.github/ci/deployments/common/k8s-relay-rollout.yaml
ensonic: ahh lines 119-121. We certainly don't need prometheus, I am not sure if we rely for k8s relay for any of the tests. We can certainly drop the copy of the prometheus-rollout.yaml, we can cleanup the k8s-relay later after maybe testing it.

drigz: The integration test tests that the relay can reach the simulated robot.

PR #108, opened 2023-03-01
Bump LLVM toolchain to Ubuntu 20.04

faximan:
16.04 is EOL. This modernizes the hermetic toolchain setup.

Thanks to some changes in how we build the corresponding RBE image, this also reduces its size >10x.


ensonic (approved):
thanks.

PR #109, opened 2023-03-02
Split the metadata handler into a separate file.

ensonic:
This is only a cleanup as a prep for future changes.


faximan (approved):
Thanks!

PR #113, opened 2023-03-02
Allow running setup-robot with human-acl@ tokens

drigz:
To reduce the privilege of the token you run setup-robot with, it's safer to give it a token for the human-acl@ SA rather than your own account. However, this currently prevents it from pulling setup-robot from a private GCR, which breaks the dev workflow. Adding this ACL fixes it.


ensonic (approved):
thanks!

PR #114, opened 2023-03-02
Also patch the default/default SA.

drigz:
A change to the order in which we update the GCR credentials introduced a bug that prevented us from running pods with images from a private GCR in the default namespace, which is triggered when setting up a new cluster pointed at a CRC dev project.

The BUILD.bazel change was required by the pre-commit hook.


ensonic (approved):
thanks

PR #116, opened 2023-03-03
Print more details on peers relying on ADCs.

ensonic:
This s a preparation to a policy on ADC usage.

### DIFF
--- a/src/go/cmd/metadata-server/metadata.go
+++ b/src/go/cmd/metadata-server/metadata.go
@@ -30,6 +30,16 @@ import (
 	"github.com/googlecloudrobotics/core/src/go/pkg/robotauth"
 	"golang.org/x/oauth2"
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	// getPodByIPRetries configures how often we retry determining the ip for a pod
+	getPodByIPRetries = 10
+	// getPodByIPWait confgured the time to sleep between the retries
+	getPodByIPWait = 500 * time.Millisecond
 )
 
 // ConstHandler serves OK responses with static body content.
@@ -50,6 +60,7 @@ type TokenHandler struct {
 	TokenSource    oauth2.TokenSource
 	Clock          func() time.Time
 	robotAuth      auth
+	k8s            *kubernetes.Clientset
 }
 
 type auth interface {
@@ -76,7 +87,7 @@ type TokenResponse struct {
 	TokenType    string `json:"token_type"`
 }
 
-func NewTokenHandler(ctx context.Context) (*TokenHandler, error) {
+func NewTokenHandler(ctx context.Context, k8s *kubernetes.Clientset) (*TokenHandler, error) {
 	_, allowedSources, err := net.ParseCIDR(*sourceCidr)
 	if err != nil {
 		return nil, fmt.Errorf("invalid source CIDR %s: %w", *sourceCidr, err)
@@ -85,6 +96,7 @@ func NewTokenHandler(ctx context.Context) (*TokenHandler, error) {
 	t := &TokenHandler{
 		AllowedSources: allowedSources,
 		Clock:          time.Now,
+		k8s:            k8s,
 	}
 	if err := t.updateRobotAuth(); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


ensonic:
New output is the part in ():
```
2023/03/03 17:13:48 Served access token to 172.26.0.39:36388 (ua="Fluent-Bit", pod="default/fluent-bit-7fbbc")
2023/03/03 17:13:48 Pod "gcscat" has no ip (yet): ""
2023/03/03 17:13:49 Retrying 1 pods without ip
2023/03/03 17:13:49 Retrying pod (kube-system/gcscat)
2023/03/03 17:13:49 Served access token to 172.26.0.252:41198 (ua="gcloud-golang/0.1", pod="kube-system/gcscat")
```
I'll eventually remove the 3 messages in the middle (retrying).

src/go/cmd/metadata-server/metadata.go
```diff
@@ -181,12 +189,85 @@ func (th *TokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		log.Printf("Failed to list namespaces: %v", err)
+	}
+	for _, ns := range nss.Items {
+		nsName := ns.ObjectMeta.Name
+		pods, err := th.k8s.CoreV1().Pods(nsName).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			log.Printf("Failed to list pods in ns=%q: %v", nsName, err)
+		}
+		for _, pod := range pods.Items {
+			// log.Printf("Search pods for %q == %q (%s/%s)", ip, pod.Status.PodIP, nsName, pod.Name)
```
mp4096, line 231: Remove?

src/go/cmd/metadata-server/metadata.go
```diff
@@ -181,12 +189,85 @@ func (th *TokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	for len(podsToRetry) > 0 && retries > 0 {
+		time.Sleep(1 * time.Second)
+		log.Printf("Retrying %d pods without ip", len(podsToRetry))
+
+		ptr := podsToRetry
+		podsToRetry = []corev1.Pod{}
+		for _, p := range ptr {
+			nsName := p.ObjectMeta.Namespace
+			podName := p.ObjectMeta.Name
+			//log.Printf("Retrying pod (%s/%s)", nsName, podName)
```
mp4096, line 253: Same

src/go/cmd/metadata-server/metadata.go
```diff
@@ -135,6 +140,9 @@ func (th *TokenHandler) NewMetadataHandler(ctx context.Context) *MetadataHandler
 	}
 }
 
+// Return an access token for the 'robot-servce' service account
```
mp4096, line 143: Typo? 
```suggestion
// Return an access token for the 'robot-service' service account
```

src/go/cmd/metadata-server/metadata.go
```diff
@@ -181,12 +189,85 @@ func (th *TokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Collect some extra data for diagnostics (aka which services rely on ADCs).
+	// User-Agent: "Fluent-Bit", "gcloud-golang/0.1"
+	ua := r.Header.Get("User-Agent")
+	pod := th.getPodNameByIP(r.Context(), ipPort[0])
```
mp4096, line 195: QQ: Assuming this gets really slow because we have many pods. How will it affect our system? Would it mean that some svcs will take longer to start up because it takes longer to get a token?

mp4096 (approved):

src/go/cmd/metadata-server/metadata.go
ensonic: The http server uses go routines. So this will not as such cause a lot of extra load, but will incur some extra latency if a service asks for ADCs right away. I don't have a better idea for this yet.

ensonic:
I added some time measurements:
* when we have an IP already it completes quickly (~40ms) - this will get slower with more namespaces and pods, so some caching as per TOD in the change might be a good idea.
* when we don't have an IP yet, it took always at least 500 ms until get got an IP, but never longer than 1 s. We can play with the 2 constants I introduced (e.g. trying 250ms sleep) if we're concerned about the 1s delay. I think in practice many service might do something else before we get here.

I'll tweak the change to add a boolean flag that is false by default. With that we can submit this, roll it out and manually enable it for testing.

src/go/cmd/metadata-server/metadata.go
```diff
@@ -30,6 +30,14 @@ import (
 	"github.com/googlecloudrobotics/core/src/go/pkg/robotauth"
 	"golang.org/x/oauth2"
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	GetPodByIPRetries = 10
```
mp4096, line 39: does this need to be exported?

src/go/cmd/metadata-server/metadata.go
```diff
@@ -30,6 +30,14 @@ import (
...
+	GetPodByIPWait    = 500 * time.Millisecond
```
mp4096, line 40: same

src/go/cmd/metadata-server/metadata.go
mp4096: Yeah, I was not concerned about system load, more about the startup latency. Most of our services create ADC-based clients on startup and do it synchronously.

I think ~1 second latency should be ok, esp. given it should not stack for the overall startup time.

+1 for the flag.

mp4096 (approved):

PR #121, opened 2023-03-21
Make the error message easier to act on

ensonic:
Only mention the relay-client in the details at the end. The most likely cause is that the client is not reachable.


faximan (approved):
Thanks!

faximan (approved):

PR #122, opened 2023-03-23
Add certificate authority selection support and Google CAS integration

### LINKED ISSUE #107: Support GCP's TLS issues in cert-manager
Since we're running by default on GCP, it would be nice to extend our cert managersetup by https://github.com/jetstack/google-cas-issuer/blob/main/README.md.

First step would be to figure out what changes we need, e.g. turn the gcloud commands from the README into terraform.
Initially we can make the GCP version the default issuer. In the future we can consider a configuration option to select the issuer (lets encrypt or gcp).

The cert-manager k8s resources are in https://github.com/googlecloudrobotics/core/tree/main/src/app_charts/base/cloud

AlejoAsd:
Closes #107.

This PR introduces the concept of Certificate Authority (CA) providers to deployment scripts. It allows selecting a certificate provider to configure Terraform files, helm charts and cluster resources necessary to generate certificates using the selected provider. 

The existing integration with Let's Encrypt has been kept as a certificate provider option. Integration with Google's Certificate Authority Service (CAS) has been added and can be selected as a certificate provider.

## Changes

1. Certificate providers. There are two options available: `lets-encrypt`, and `google-cas`. The default option was set to `lets-encrypt` for backwards compatibility purposes.
2. Some configuration variables have been added. Existing configurations can be updated by running `deploy.sh set-config`.

## How to test

1. Run `./deploy.sh set-config <PROJECT_ID>` to configure new environment variables.
    * Select `google-cas` as the certificate provider.
    * Fill in the Common Name (`CN`) and Organization (`O`) certificate fields. Optionally, fill in the Organizational Unit (`OU`) certificate field.
2. Run `./deploy.sh update <PROJECT_ID>` to deploy the configuration.
3. Verify that the `cloud-robotics` certificate was issued correctly.

The output should look something like this

```bash
# Make sure kubectl is using the correct config and context
$ kubectl get certificate cloud-robotics 
NAME             READY   SECRET               AGE
cloud-robotics   True    cloud-robotics-tls   XXm
```

4. Verify that `www.endpoints.robco-ensonic.cloud.goog` is using the certificate.

```sh
# Make sure to set <PROJECT_ID>
openssl s_client -showcerts -connect www.endpoints.$PROJECT_ID.cloud.goog:443
```


AlejoAsd:
There is currently a configuration mismatch between the `cert-manager` and `google-cas-issuer` k8s service accounts that prevents the certificates from being approved. Currently working on fixing this. This is still in a good place to have a first review.

AlejoAsd:
Some additional fixes have been put in place.

I have tested creating certificates both with the `lets-encrypt` and `google-cas` certificate providers and they are working as intended. The infrastructure and cluster resources are correctly created and configured.

It is important to note that because Google Certificate Authority System is a private CA, it is not recognized by default by browsers (returns `NET::ERR_CERT_AUTHORITY_INVALID`) and needs to be added to the list of valid CAs. This might have additional considerations for the implementation of the `google-cas` provider.

src/app_charts/base/cert-manager-google-cas-issuer-cloud.values.yaml
```diff
@@ -0,0 +1,19 @@
...
+# Reference: https://github.com/jetstack/google-cas-issuer/blob/main/deploy/charts/google-cas-issuer/values.yaml
+
+# No values are required for now.
+# This was put in place to add values in the future without requiring additional configuration in other files.
+# If values are added this disclaimer should be removed.
+
+# The Kubernetes service account must be annotated in order to impersonate a GCP service account using workload identity.
+serviceAccount:
+  annotations:
+    iam.gke.io/gcp-service-account: sa-google-cas-issuer@robco-ensonic.iam.gserviceaccount.com
```
ensonic, line 11: we can't hardcode this.

ensonic:
Thanks a lot. Super cool! WHich configuration did you currently deploy in the testing project (self signed?)
A few comments wrt. to the PR description:

Changes:
Please keep "lets-encrypt" as a default for backwards compatibility?

How to test:
1 configure new environment variables:
Can we use PROJECT_ID as defaults for CN and O?
4 verify:
We can propose to run:
```
openssl s_client -showcerts -connect www.endpoints.$PROJECT_ID.cloud.goog:443
```
(Also please edit and use PROJETC_ID consistently).

ensonic:
> Some additional fixes have been put in place.
> 
> I have tested creating certificates both with the `lets-encrypt` and `google-cas` certificate providers and they are working as intended. The infrastructure and cluster resources are correctly created and configured.
> 
> It is important to note that because Google Certificate Authority System is a private CA, it is not recognized by default by browsers (returns `NET::ERR_CERT_AUTHORITY_INVALID`) and needs to be added to the list of valid CAs. This might have additional considerations for the implementation of the `google-cas` provider for both browsers and robots.

That is a huge bummer and totally unexpected. I will reach out to the internal team that requested this change.

AlejoAsd:
Thanks for the review! I will address the comments on Monday.

As you say, Google does provide public certs, but they are part of another
service that is not managed by cert-manager.

From the little time I spent looking into it the recommended way to
configure public certs in GKE is through ingress annotations, which is a
bit cumbersome given our nginx setup. I will keep digging to see if there a
better way to tackle this.

src/app_charts/base/cert-manager-google-cas-issuer-cloud.values.yaml
AlejoAsd: Addressed in 9f42587.

AlejoAsd:
All comments have been addressed. If you're okay with it, I'd like to merge this and include support for public Google certs service in another PR.

ensonic (approved):
Thanks a lot!

PR #123, opened 2023-03-25
Adding akri to support usb devices

jainajinkya:
Adding helm charts from third-party library [project-akri](https://github.com/project-akri/akri) to support pluggable usb devices.


src/app_charts/base/BUILD.bazel
```diff
@@ -30,16 +30,28 @@ helm_template(
     values = "cert-manager-robot.values.yaml",
 )
 
+helm_template(
+    name = "akri-chart.robot",
+    chart = "//third_party/akri:Chart.yaml",
```
ensonic, line 35: this need to be the chart's tgz

src/app_charts/base/BUILD.bazel
```diff
@@ -52,6 +64,7 @@ app_chart(
     visibility = ["//visibility:public"],
 )
 
+
```
ensonic, line 67: nit: remove

third_party/akri/BUILD.bazel
ensonic, line 1: Could you add a README:md next to this (or add comments in to the BUILD.bazel files) that describe how the files here where obtained /produced. This is crucial to ensure we can update things.

third_party/akri/README.md
```diff
@@ -0,0 +1,24 @@
+Helm chart for akri (version 0.8.23) included here was pulled from the project-akri github repo using the following steps and modified to be compatible with the Intrinsic codebase:
+```
+helm repo add akri-helm-charts https://project-akri.github.io/akri/ 
+helm fetch  akri-helm-charts/akri --untar
+```
+
+Following modifications were done to the "vanilla" helm chart pulled from the
```
ensonic, line 7: could we store patch files in third_party/akri, so that one can download a new version and apply the patch?
the patch files would also better document what exactly was changed and the instructions here would be simplified by a one line to apply the patches.

ensonic (commented):
The last part we need to figure out is how to not enable this by default everywhere. For things like fluentbit we did this:
https://github.com/googlecloudrobotics/core/blob/main/src/app_charts/base/robot/fluent-bit.yaml

Maybe we can patch the third_party/akri/values.yaml to flip the relevant xx/enable fields. I'll brainstorm with my colleagues on alternative ideas.

third_party/akri/README.md
ensonic: Another idea to make this optional would be to store it as an cloud-robotics-app like app_charts/prometheus, so that users can define an app-rollout if they want to use it. That'd be my current preference & sorry for the back and forth.

src/app_charts/base/BUILD.bazel
ensonic, line 1: Can we try adding src/app-charts/akri/ and making the change you made in this file there instead. You can take src/app_charts/prometheus/BUILD.bazel as an example. In your case omit all the cloud parts.

third_party/akri/README.md
jainajinkya: Sounds good. Updated.

src/app_charts/base/BUILD.bazel
jainajinkya: Updated.

third_party/akri/update-akri.sh
```diff
@@ -0,0 +1,52 @@
...
+# fetch the latest version
+helm repo add akri-helm-charts https://project-akri.github.io/akri/ 
+helm repo update akri-helm-charts
+helm pull akri-helm-charts/akri --untar
+
+# Update crds
+cp -r ./akri/crds/* .
+
+# Edit helm-charts
+sed -i 's/v2/v1/' ./akri/Chart.yaml
```
ensonic, line 12: the helm_template rule has helm_version arg, where you can select '3'. If that is watch you are patching here.

jainajinkya: In `akri/Chart.yaml` the apiVersion is mentioned as `v2`. However, our build rule enforces that the apiVersion must be `v1`. This line patches the apiVersion to be `v1`.

jainajinkya: Updated.

ensonic (approved):
Thanks!

PR #126, opened 2023-03-31
Added support for multi-line parsing of Abseil logs.

hauke76:


hauke76:
I checked this change by running:

helm template fluent-bit fluent/fluent-bit --version ${TEMPLATE_VERSION} -f fluent-bit-values.yaml --skip-tests > ${OUTPUT}
kubectl --context minikube apply -f ~/tmp/fluent-bit.yaml

I then went to k9s and restarted a POD with multiline logs and checked on Pantheon if they were correctly incorporated.

drigz:
@faximan As he set something similar up for fluentd.

ensonic (approved):
Thanks!

faximan:
> @faximan As he set something similar up for fluentd.

I think the relevant commit is https://github.com/googlecloudrobotics/core/commit/ed68c64c1b2deafa33a0fb12da7a29a10a8adfff. It looks like Hauke is matching `[\s\S]*` instead of `(.|\n)*` as well as using a different regexp to detect a new line start. I have no idea if one is better than the other :)

faximan (approved):

hauke76:
Let me try to explain the three regexs:
1) start: ^((W|I|E|F))([0-9]{4}) ([^ ]+)\s+([-0-9]+) (\S+:\d+)] (.*)$
2) cont: ^(?!(((W|I|E|F))([0-9]{4}) ([^ ]+)\s+([-0-9]+) (\S+:\d+)]))(.*)$
3) blocks: ^(?<severity>(W|I|E|F))([0-9]{4}) (?<time>[^ ]+)\s+(?<pid>[-0-9]+) (?<source>\S+:\d+)] (?<message>[\s\S]*)$

1) start: identifies the first log row of a multi-line log entry; matches only a single row
2) cont: uses a negative look ahead to identify any row which does not match 1); matches only a single row
3) blocks: matches the Abseil preamble and then any whitespace and non-whitespace character including \n and thus multiple rows

The diff ([\s\S]*) vs. ((.|\n)*.) seems very similar [\s\S]* == (.|\n)* and I believe the additional . which matches again any character except line terminators is unnecessary in the latter.

All in all, I think this is pretty similar. I verified all this and pulled the explanations from regex101.com.

PR #127, opened 2023-04-04
LLVM 14 and glibc 2.24

faximan:
This corresponds to cl/518900665 and cl/519950444, aligning internal with external.


ensonic (approved):
Thanks!

PR #131, opened 2023-04-11
Move master-host to an annotation

drigz:
This allows for longer hostnames, in particular those of GCE instances. This doesn't delete the old label, because that's hard with the default "merge labels" behavior of CreateOrUpdateRobot(), so it's easier to manually migrate the labels.

@Ongy


ensonic (approved):
Thanks!

PR #132, opened 2023-04-12
Cleanup IoT registry on deploy.

csieber:
If terraform apply fails if the IoT registry still has registered devices.


deploy.sh
```diff
@@ -235,6 +237,27 @@ function cleanup_helm_data {
   kc -n kube-system delete cm -l OWNER=TILLER 2> /dev/null || true
 }
 
+function cleanup_iot_devices {
```
ensonic, line 240: should we add an early return, a la
gcloud services list --project="${GC_PROJECT_ID}" | grep -q cloudiot.googleapis.com || return

csieber: The service list is probably slower than the quick fail of the next command :)

csieber: I will test it

ensonic (approved):
thanks

PR #133, opened 2023-04-13
Fix akri-app.

ensonic:
* Add subdir for robot deploymnet. Without it we never pull the workloads.
* Remove curly braces from the akri-crd. These cause errors when applying this as part of a chart.
* Simply the update script. We'll do customization via values.yaml


third_party/akri/update-akri.sh
```diff
@@ -3,48 +3,9 @@
 # fetch the latest version
 helm repo add akri-helm-charts https://project-akri.github.io/akri/ 
 helm repo update akri-helm-charts
-helm pull akri-helm-charts/akri --untar
+helm pull akri-helm-charts/akri
```
csieber, line 6: would it make sense to pin a version here?

third_party/akri/update-akri.sh
```diff
@@ -3,48 +3,9 @@
...
-# opcua - May require more updates
-sed -i 's/.Chart.AppVersion/.Values.akriAppVersion/g' ./akri/templates/opcua-discovery-handler.yaml
-
-# Repackage updated helm-charts 
-helm package ./akri
-
-# Cleanup
-rm -rf ./akri
-
+tar xzf akri-*.tgz --strip-components=2 akri/crds/
```
csieber, line 9: if the old tgz is not deleted, wouldn't this result in the old and new version both being extracted?

third_party/akri/update-akri.sh
ensonic: Good idea. Done.

third_party/akri/update-akri.sh
ensonic: Using the version here now.

Issue #130, opened 2023-04-11
Try porting from rules_docker to rules_oci


ensonic:
In https://github.com/bazelbuild/rules_docker/issues/2204#issuecomment-1465317395 Alex Eage suggested to use https://github.com/bazel-contrib/rules_oci. A quick `git grep rules_docker` looks like it is not too much of a change (15 files).

ensonic:
Fixed. Thanks @therjak

PR #136, opened 2023-04-19
Get base images for go builds independent of rules_docker version

AlejoAsd:
Closes issue #130.

This PR alters the bazel build to use `rules_oci` to pull go base images using a version that's decoupled for the `rules_docker` version. This allows for us to update the base image version without having to depend on `rules_docker`.

The latest version of both the `base` and `static` distroless version was used to avoid unexpected breakages. 

Note that `oci_rules` `pull` supports using the `latest` tag. The macro that pulls base images can be configured to always pull the latest images and be up to date if we are okay with assuming the potential upstream breakage risks it introduces.

### DIFF
--- a/bazel/base_images.bzl
+++ b/bazel/base_images.bzl
@@ -1,0 +1,27 @@
+load("@io_bazel_rules_docker//repositories:go_repositories.bzl", _go_deps = "go_deps")
+load("@rules_oci//oci:pull.bzl", "oci_pull")
+
+def go_base_images():
+    # This call is technically optional but is included here for safety as it is idempotent and required,
+    _go_deps()
+
+    oci_pull(
+        name = "go_image_base",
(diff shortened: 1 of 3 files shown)


bazel/base_images.bzl
```diff
@@ -0,0 +1,27 @@
+load("@io_bazel_rules_docker//repositories:go_repositories.bzl", _go_deps = "go_deps")
+load("@rules_oci//oci:pull.bzl", "oci_pull")
+
+def go_base_images():
+    # This call is technically optional but is included here for safety as it is idempotent and required,
+    _go_deps()
+
+    oci_pull(
```
ensonic, line 8: could you add a comment on how you generated the sha256: values and how one would update them?

https://github.com/GoogleContainerTools/distroless is not doing releases.

ensonic (approved):
Thanks!

PR #138, opened 2023-04-24
Fix CI for external contributors & bots

faximan:
External contributors don't have access to GitHub repo secretes, by design, or else they could easily exfiltrate them. This caused us to pass an invalid credentials file to Bazel and Bazel to crash with a cryptic error 37 on startup for such RPs.

I could not find exactly what "secret" will be passed in such cases (empty variable or an error?) so try to parse the resulting file as JSON as proxy.

For external PRs, simply run presubmit testing without RBE. It will be a bit slower (the test currently takes [~20 mins](https://github.com/googlecloudrobotics/core/actions/runs/4787083927/jobs/8511852870?pr=138)), but it will work.


ensonic (approved):
Thanks!

PR #139, opened 2023-04-25
Fix missing rename

faximan:
This file was renamed in https://github.com/googlecloudrobotics/core/pull/138 but I forgot to update this reference.


ensonic (approved):
thanks

PR #141, opened 2023-05-09
Avoid silent hang when failing to patch SA

drigz:
Investigating b/281647304, I've observed some gcr-credential-refresher instances that hang forever, logging nothing. The stacktrace points to this retry loop, so adding a limit should prevent the hang from breaking the gcr-credential-refresher, plus give us more info on the root cause.


ensonic (approved):
Thanks

therjak (approved):

PR #147, opened 2023-05-23
Update Go base images

AlejoAsd:
Base Go images have been updated to use the latest versions. Images for `gcr.io/distroless/base` and `gcr.io/distroless/static` have been updated.


ensonic (approved):
Thanks!

PR #148, opened 2023-05-23
fluent-bit: add kernel log

therjak:
this uses the systemd plugin to get around access problems of the kmsg plugin

### DIFF
--- a/src/app_charts/base/fluent-bit-values.yaml
+++ b/src/app_charts/base/fluent-bit-values.yaml
@@ -78,6 +78,12 @@ config:
         Read_From_Tail On
 
     [INPUT]
+        Name systemd
+        Tag kernel.${MY_NODE_NAME}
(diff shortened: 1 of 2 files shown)


ensonic:
I think you need to run src/app_charts/base/fluent-bit-helm.sh and include the changed file

therjak:
was I not the first to forget to run src/app_charts/base/fluent-bit-helm.sh? most of the additional changes seem unrelated to my change.

methylDragon:
> was I not the first to forget to run src/app_charts/base/fluent-bit-helm.sh? most of the additional changes seem unrelated to my change.

Yeah that's what happened to me too 😬 , I think it's alright.
You might want to strip the whitespace that got added. If you're on VS Code you can `Trim Trailing Whitespace` in the command palette

drigz (approved):

PR #156, opened 2023-06-09
Update ingress-nginx to 1.8.0

drigz:
This pulls in an Alpine update which should fix some CVE warnings. The newer release is incompatible with k8s 1.23 but GKE has auto-updated to 1.24 everywhere now 1.23 is EOL.

I changed the RBAC policy based on the v1.3.0 release notes (needs permission for coordination.k8s.io/leases). Don't ask me why upstream added it to both the Role and the ClusterRole.
https://github.com/kubernetes/ingress-nginx/releases/tag/controller-v1.3.0

Tested by deploying to robco-rodrigoq and checking the nginx dashboards in Grafana.

### DIFF
--- a/METADATA
+++ b/METADATA
@@ -4,7 +4,7 @@ third_party {
   # https://nvd.nist.gov/products/cpe/search
   security {
     tag: "NVD-CPE2.3:cpe:/a:grafana:grafana:9.1.17"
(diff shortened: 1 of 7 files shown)


ensonic (approved):
Thanks!

PR #161, opened 2023-06-12
Update cert-manager to v1.12.1

drigz:
From the release notes, there are no breaking changes that should affect us. Tested on robco-rodrigoq by checking logs for any errors (it just reused the existing certs so I couldn't check it fully).


ensonic (approved):
Lets update all of it to giza-workcells and xfa-awesome-alpha next week to give it some testing. I'll bump oauth2-proxy tomorrow.

PR #163, opened 2023-06-19
WIP: switch from rules_docker to rules_oci

therjak:


bazel/build_rules/app_chart/push_all.bzl
```diff
@@ -69,7 +69,6 @@ def push_all(name, images = {}, **kwargs):
         push_targets.append(push_target)
         container_push(
             name = push_target,
-            format = "Docker",
```
ensonic, line 72: We need to change line 23 and replace --dst wth --repository and --tag

deploy.sh
```diff
@@ -118,7 +118,7 @@ function terraform_init {
   fi
 
   local ROBOT_IMAGE_DIGEST
-  ROBOT_IMAGE_DIGEST=$(cat bazel-bin/src/go/cmd/setup-robot/setup-robot.push.digest)
+  ROBOT_IMAGE_DIGEST=$(cat bazel-bin/setup-robot.digest)
```
ensonic, line 121: Also need to change line 86 and replace --dst with --repository and --tag. Same for .github/ci/common.sh:99

ensonic:
I've got this error: https://gist.github.com/ensonic/9560e2a058ff30fa0cdfed36304938e3
Lokks like it is still using the rules-docker for the setup-robot container?

ensonic:
If it is hard to fix we can use `bazel run src/go/cmd/setup-robot:setup-robot.push --` instead of invoking the two push scripts in deploy.sh

ensonic:
This is work now for me. I also did a 2nd run after bazel clean --expunge.

PR #166, opened 2023-07-03
Extend the monitoring of the relay to also track the backend timing.

ensonic:
This lets us observe the overall duration, the backend duration and as a delta (overall-backend) the overhead. Especially the overhead is hopefully useful to spot infrastructure issues on the relay side.

### DIFF
--- a/src/go/cmd/http-relay-server/broker.go
+++ b/src/go/cmd/http-relay-server/broker.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


src/proto/http-relay/http_over_rpc.proto
```diff
@@ -46,4 +46,5 @@ message HttpResponse {
   optional bytes body = 3;
   optional bool eof = 5;
   repeated HttpHeader trailer = 6;
+  optional int64 backend_duration_ms=7;
```
therjak, line 49: should this be a google/protobuf/duration.proto ?
https://pkg.go.dev/google.golang.org/protobuf/types/known/durationpb

src/go/cmd/http-relay-server/broker.go
```diff
@@ -209,13 +232,22 @@ func (r *broker) SendResponse(resp *pb.HttpResponse) error {
...
 	pr.responseStream <- resp
 	r.m.Unlock()
-	brokerRequests.WithLabelValues("server_response", backendName).Inc()
-	brokerResponseDurations.WithLabelValues("server_response", backendName).Observe(duration)
-	log.Printf("[%s] Delivered response to client (%d bytes), elapsed %.3fs", id, len(resp.Body), duration)
+	brokerRequests.WithLabelValues("server_response", backendName, pr.requestPath).Inc()
+	brokerResponseDurations.WithLabelValues("server_response", backendName, pr.requestPath).Observe(duration)
 	if resp.GetEof() {
 		close(pr.responseStream)
+		backendDuration := (time.Duration(resp.GetBackendDurationMs()) * time.Millisecond).Seconds()
```
therjak, line 239: this first creates a Duration in a wrong unit (BackendDurationMs() is in Milisecond but Duration is in ns) and fixes it afterwards. please get the unit right before converting to time.Duration.

src/proto/http-relay/http_over_rpc.proto
ensonic: technically yes, but I don't want to add the extra dependency here. Do you think it is worth it?

src/go/cmd/http-relay-server/broker.go
ensonic: Yes, that looks weird, but I can't write:
```
time.Duration(resp.GetBackendDurationMs() * time.Millisecond))
```
as the compiler complains about:
```
invalid operation: resp.GetBackendDurationMs() * time.Millisecond (mismatched types int64 and time.Duration)
```
Any better idea?`

therjak (approved):

ensonic:
Thanks!

PR #172, opened 2023-07-14
Added tracing to the relay client and server.

hauke76:
This adds trace support to our relay server and client to be able to better debug and monitor its performance.

Traces are logged when they are enabled through the TRSID cookie.

The generated traces (see image below) look at times strange. My working hypothesis is that the relay server sends its reply to the user client (see `server.userClientRequest()`) before actually finishing the async receiving of the response in `server.serverResponse()`.

image


src/go/cmd/http-relay-client/main.go
```diff
@@ -152,35 +163,31 @@ func marshalHeader(h *http.Header) []*pb.HttpHeader {
...
+		header.Add(*h.Name, *h.Value)
+	}
+}
+
+func createBackendRequest(breq *pb.HttpRequest) (*http.Request, error) {
 	id := *breq.Id
 	targetUrl, err := url.Parse(*breq.Url)
-	if err != nil {
-		return nil, nil, err
-	}
```
ensonic, line 165: why this change?

src/go/cmd/http-relay-client/main.go
```diff
@@ -190,10 +197,31 @@ func makeBackendRequest(local *http.Client, breq *pb.HttpRequest) (*pb.HttpRespo
 		log.Printf("%s", dump)
 	}
 
+	return req, nil
+}
+
+// This function builds and executes a http.Request from the proto request we received from the web-client.
```
ensonic, line 203: "user-client" or maybe end-user client. here and below

src/go/cmd/http-relay-client/main.go
```diff
@@ -449,10 +511,16 @@ func handleRequest(remote *http.Client, local *http.Client, req *pb.HttpRequest)
...
 					resp.BackendDurationMs = proto.Int64(duration.Milliseconds())
 					// see makeBackendRequest()
-					urlPath := strings.TrimPrefix(*req.Url, "http://invalid")
+					urlPath := strings.TrimPrefix(*pbreq.Url, "http://invalid")
 					log.Printf("[%s] Backend request duration: %.3fs (for %s)", *resp.Id, duration.Seconds(), urlPath)
+				} else {
+					// Q(hauke): When are we ending up in this branch?
+					// What are the semantics and why are we not setting a request duration?
+					// Even in a streaming case I would expect a duration which represents the
+					// processing time of the last item.
```
ensonic, line 520: we could track each backend request and report the progress. let try as a followup

src/go/cmd/http-relay-client/main.go
```diff
@@ -449,10 +511,16 @@ func handleRequest(remote *http.Client, local *http.Client, req *pb.HttpRequest)
...
 				}
-				return postResponse(remote, resp)
+				err = postResponse(remote, resp)
+				return err
```
ensonic, line 523: revert?

src/go/cmd/http-relay-client/main.go
```diff
@@ -563,7 +645,7 @@ func main() {
 			}
 		}
 
-		transport = h2transport
+		transport = &ochttp.Transport{Base: h2transport}
```
ensonic, line 648: move `&ochttp.Transport{Base: ` to where we set the transport: `Transport: transport,`

src/go/cmd/http-relay-server/broker.go
```diff
@@ -127,6 +132,7 @@ func (r *broker) RelayRequest(server string, request *pb.HttpRequest) (<-chan *p
 		r.req[server] = make(chan *pb.HttpRequest)
 	}
 	if r.resp[id] != nil {
+		r.m.Unlock()
```
ensonic, line 135: wow, could you send this as a separate PR (you can send a PR right from the github webpage).

src/go/cmd/http-relay-server/broker.go
```diff
@@ -212,7 +221,7 @@ func (r *broker) PutRequestStream(id string, data []byte) bool {
 	return true
 }
 
-// SendResponse delivers the HttpResponse to the client handler that created the
+// SendResponse delivers the HttpResponse to the web-client handler that created the
```
ensonic, line 224: again, user-client?

ensonic (commented):
Thanks!

Do we need a way to configure the sampling rate?

src/go/cmd/http-relay-server/broker.go
hauke76: Found a few more places and fixed them too.

src/go/cmd/http-relay-server/broker.go
hauke76: Fixed separately in #173

hauke76:
According to Christian, we don't need to specify the sample rate as it will be propagated. Another question which I would like to address in a follow up PR is whether we enable tracing by default. Christian added it as a safety guard to prevent excessive cost generation.

ensonic (approved):
thanks!

PR #177, opened 2023-07-19
apps: allow versioned inline apps

coconutruben:
This change enables inline apps to be versioned as well.
It does so by adopting a label and naming convention,
and using the Version attribute in the Robot selector of
AppRollouts to fetch a different App depending on the version.
It finds those versioned apps using labels, and has a validator to
ensure app names and labels don't drift too far from each other.

Lastly, this change adds unit tests to test the new behavior.

Note: this removes a check in the chartassignment validator, namely that Inline
can now coexist with Version.

TEST=go test ./...
TEST=deploy to intrinsic-coconutruben

### DIFF
--- a/src/app_charts/base/cloud/app-management.yaml
+++ b/src/app_charts/base/cloud/app-management.yaml
@@ -79,6 +79,33 @@ spec:
     name: cluster-authority
 ---
 apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: app-rollout-controller
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/app-rollout-controller
+webhooks:
+- name: apps.apps.cloudrobotics.com
+  admissionReviewVersions: ["v1"]
+  failurePolicy: Fail
+  clientConfig:
+    service:
+      namespace: {{ .Release.Namespace }}
+      name: app-rollout-controller
+      path: /app/mutate
+  rules:
+  - apiGroups:
+    - apps.cloudrobotics.com
+    apiVersions:
+    - v1alpha1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - apps
+  sideEffects: None
+---
+apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: app-rollout-controller
--- a/src/app_charts/base/cloud/app-management.yaml
+++ b/src/app_charts/base/cloud/app-management.yaml
@@ -104,14 +104,6 @@ webhooks:
     resources:
     - apps
   sideEffects: None
----
-apiVersion: admissionregistration.k8s.io/v1
-kind: ValidatingWebhookConfiguration
-metadata:
-  name: app-rollout-controller
-  annotations:
-    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/app-rollout-controller
-webhooks:
 - name: approllouts.apps.cloudrobotics.com
   admissionReviewVersions: ["v1"]
   failurePolicy: Fail
--- a/src/app_charts/base/cloud/app-management.yaml
+++ b/src/app_charts/base/cloud/app-management.yaml
@@ -79,7 +79,7 @@ spec:
     name: cluster-authority
 ---
 apiVersion: admissionregistration.k8s.io/v1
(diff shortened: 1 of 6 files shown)


coconutruben:
We don't have to do it this way, this is just the current thing I tried out, and tested, very open to feedback here

drigz (commented):
Drive-by comment, leaving rest of review to Stefan since you already discussed the change with him. Stefan, feel free to reassign to me if you have too many reviews.

src/go/pkg/controller/chartassignment/validator.go
```diff
@@ -80,7 +80,7 @@ func (v *chartAssignmentValidator) validate(cur, old *apps.ChartAssignment) erro
 	}
 	c := cur.Spec.Chart
 	if c.Inline != "" {
-		if c.Repository != "" || c.Name != "" || c.Version != "" {
+		if c.Repository != "" || c.Name != "" {
 			return fmt.Errorf("chart repository, name, and version must be empty for inline charts")
```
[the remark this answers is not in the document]
coconutruben: thanks, updated!

src/app_charts/base/cloud/app-management.yaml
```diff
@@ -85,6 +85,33 @@ metadata:
...
+    - CREATE
+    - UPDATE
+    resources:
+    - apps
+  sideEffects: None
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: app-rollout-controller
```
ensonic, line 111: I don't think you can have this twice, but the content of webhooks is a list, so maybe you can merge this?

coconutruben: Thanks for pointing that out, even though it built, it was getting overwritten. And there was a fixup needed for the hook implementation itself. I merged them now
```
NAME                     WEBHOOKS   AGE
app-rollout-controller   2          7d5h
```

src/go/pkg/controller/approllout/controller.go
```diff
@@ -411,23 +446,49 @@ func (r errRobotSelectorOverlap) Error() string {
...
+		robots, err := matchingRobots(robots, rcomp.Selector)
 		if err != nil {
 			return nil, errors.Wrap(err, "select robots")
 		}
+		// map is populated by for all the rcomp.Version, no need to check ok
+		app, ok := appVersions[rcomp.Version]
+		if !ok {
+			return nil, fmt.Errorf("no App %q (Version: %q) found", rollout.Spec.AppName, rcomp.Version)
+		}
+		comps := app.Spec.Components
```
ensonic, line 491: I don't think this survices the scope of the loop. See eg. https://stackoverflow.com/questions/33823209/variable-in-for-loop-undefined-outside

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -319,6 +321,174 @@ spec:
 	}
 }
 
+// generateApp will generate an app for testing
+func generateApp(name, version, robotPayload, cloudPayload string) apps.App {
+	app := apps.App{}
```
ensonic, line 326: can you write:
app := apps.App{
  Name: name,
  Labels: map[string]string{
    ..
  },
}

}

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -319,6 +321,174 @@ spec:
...
+	app.Spec.Components.Cloud.Name = "cloud"
+	app.Spec.Components.Cloud.Inline = cloudPayload
+	app.Spec.Components.Robot.Name = "robot"
+	app.Spec.Components.Robot.Inline = robotPayload
+	return app
+}
+
+// generateRobot will generate a robot named |name| with the labels
+func generateRobot(name string, labels map[string]string) registry.Robot {
+	robot := registry.Robot{}
```
ensonic, line 341: same here and below

ensonic (approved):
Please also sent a followup to update
https://github.com/googlecloudrobotics/core/blob/main/docs/concepts/app-management.md

src/go/pkg/controller/approllout/controller_test.go
coconutruben: thanks, good catch!

coconutruben:
> Please also sent a followup to update https://github.com/googlecloudrobotics/core/blob/main/docs/concepts/app-management.md

coming in another PR, thank you!

PR #179, opened 2023-07-20
Update rules_go and Go dependencies

drigz:
This required a slightly tricky manual change to go_repositories.bzl: https://github.com/bazelbuild/rules_go/issues/3625

I also updated the Go dependencies trying to fix this, which didn't seem to be necessary but is probably worth doing regardless.


therjak (approved):

ensonic (approved):
thanks

PR #181, opened 2023-07-20
Mocked time.Since in the relay client to fix flaky unit tests.

### LINKED ISSUE #180: Flaky test: //src/go/cmd/http-relay-client:go_default_test
```
bazel test --runs_per_test=1000 //src/go/cmd/http-relay-client:go_default_test
[...]
//src/go/cmd/http-relay-client:go_default_test                           FAILED in 14 out of 1000 in 20.1s
  Stats over 1000 runs: max = 20.1s, min = 0.1s, avg = 0.3s, dev = 1.3s
```

First observed in presubmit at https://source.cloud.google.com/results/invocations/f2b19b1d-478e-4a1b-a324-a420f7f0ff3c/targets/%2F%2Fsrc%2Fgo%2Fcmd%2Fhttp-relay-client:go_default_test/tests

hauke76:
This ensures that we can alway return 0 ms elapsed time in a unit test. Fixes #180.

### DIFF
--- a/src/go/cmd/http-relay-client/main.go
+++ b/src/go/cmd/http-relay-client/main.go
@@ -115,6 +115,11 @@ var (
 	ErrForbidden = errors.New(http.StatusText(http.StatusForbidden))
 )
 
+// This is a package internal variable which we define to be able to overwrite
+// the measured time during unit tests. This is a light weight alternative
+// to mocking the entire time interface and passing it along all call paths.
+var timeSince = time.Since
+
 func addServiceName(span *trace.Span) {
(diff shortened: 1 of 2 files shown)


ensonic (approved):
Thanks. I guess this makes comparing the body in the response simpler.

hauke76:
Before this change, the unit test failed if the relay-server response took longer than a millisecond. It did a hard check on `BackendDurationMs: proto.Int64(0)` in the request body proto the client sent as a response to the relay server. With this change, we can ensure that it stays zero or actually any other value we want.

hauke76:
I have not found a solution in gock to partially match a body which also seems hard since we have a binary body if I am not mistaken.

PR #183, opened 2023-07-20
generate.sh

coconutruben:
This change modifies generate.sh to work again. This is required to keep
our generated code for CRDs in sync.

Note: there is newer code generation that introduces breaking changes however. This is why the release is kept the same as before, and just the tooling updated.

BUG=NONE
TEST=./generate.sh
     cd ../..
     deploy.sh update intrinsic-coconutruben


drigz (approved):
Many thanks! That is a big simplification.

ensonic (approved):
Nice. Thanks!

PR #187, opened 2023-07-25
More fixes post OCI migration

faximan:
I missed this in https://github.com/googlecloudrobotics/core/pull/186.

Now tested by actually running this part of the code too: https://github.com/googlecloudrobotics/core/actions/runs/5655155089
https://github.com/googlecloudrobotics/core/actions/runs/5655020052

How we read the setup robot digest also needed to change, as the `.json` was not part of the binary distribution. I manually checked that `bazel-bin/src/bootstrap/cloud/setup-robot.digest` and `src/go/cmd/setup-robot/setup-robot-image/index.json` contains the same digest.

### DIFF
--- a/.github/ci/common.sh
+++ b/.github/ci/common.sh
@@ -86,13 +86,19 @@ function release_binary {
       //src/app_charts:push \
       //src/go/cmd/setup-robot:setup-robot.push
 
+  # The push scripts depends on binaries in the runfiles.
+  local oldPwd=$(pwd)
   # The tag variable must be called 'TAG', see cloud-robotics/bazel/container_push.bzl
   for t in latest ${DOCKER_TAG}; do
-    bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh \
+    cd ${oldPwd}/bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh.runfiles/cloud_robotics
+    ${oldPwd}/bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh \
       --repository="${CLOUD_ROBOTICS_CONTAINER_REGISTRY}/setup-robot" \
       --tag="${t}"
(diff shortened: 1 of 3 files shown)


.github/ci/common.sh
```diff
@@ -86,13 +86,17 @@ function release_binary {
...
+  # The push script depends on binaries in the runfiles.
+  local oldPwd=$(pwd)
+  cd bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh.runfiles/cloud_robotics
   # The tag variable must be called 'TAG', see cloud-robotics/bazel/container_push.bzl
   for t in latest ${DOCKER_TAG}; do
-    bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh \
+    ${oldPwd}/bazel-bin/src/go/cmd/setup-robot/push_setup-robot.push.sh \
       --repository="${CLOUD_ROBOTICS_CONTAINER_REGISTRY}/setup-robot" \
       --tag="${t}"
     TAG="$t" bazel-bin/src/app_charts/push "${CLOUD_ROBOTICS_CONTAINER_REGISTRY}"
```
therjak, line 97: here you need ${oldPwd}/ as well

.github/ci/common.sh
```diff
@@ -86,13 +86,19 @@ function release_binary {
       //src/app_charts:push \
       //src/go/cmd/setup-robot:setup-robot.push
 
+  # The push scripts depends on binaries in the runfiles.
+  local oldPwd=$(pwd)
```
therjak, line 90: there was a reason to write this in two lines. I think something related to how errors are handled in bash. (Is it possible to get the review comments of my cl? I think there was a reference to the reason)

faximan: I think the conversation history should still be there, but I cannot find any such thread on your PR. But in any case, I think you are right (there is a g3 linter often complaining about this), so I updated this.

therjak (approved):

PR #190, opened 2023-07-31
Use gke-gcloud-auth-plugin in relay test

drigz:
This requires some hacks to make gcloud happy in the test environment. I confirmed that it fails with `TEST FAILED: id command did not run, output was ""`.

b/293827843


hauke76:
Thanks a lot for addressing this so quickly!

hauke76 (approved):

PR #195, opened 2023-08-07
DONOTSUBMIT: WIP

hauke76:
In the past, we used the default of 2 idle connections per host which (almost entirely) prevented connection reuse between the relay-server and relay-client.


ensonic:
A releayed machine is usually only used by a single user at a time and the relay clients there will have a single host as a peer (the matching, per service,  relay server). 
Do we have evidence that more per-host idle connections would help?

My understanding is that an idle connection would speed-up a new connection request (since we still have one). Since we have multiple connections in flight already, this would only speed-up enqueuing a new connection, which in turn would only help if we run out of enqueued connections.

If that is the case I would suggest to try adding some variation on the timeout (so that we don't enqueue N connections and they run into timeout at the same time).

hauke76:
If you read through this, it has a relatively good explanation: http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

The most important paragraphs are:
"In the above example, there are 100 goroutines trying to concurrently make requests to the same host, but the connection pool can only hold 2 sockets. So in the first “round” of the goroutines finishing their http request, 2 of the sockets will remain open in the pool, while the remaining 98 connections will be closed and end up in the TIME_WAIT state.

Since this is happening in a loop, you will quickly accumulate thousands or tens of thousands of connections in the TIME_WAIT state. Eventually, for that particular host at least, you will run out of ephemeral ports and not be able to open new client connections. For a load testing tool, this is bad news."

I actually ran tests which confirmed the behavior.

PR #197, opened 2023-08-07
Factored out the relay server into a separate class.

hauke76:
This CL does not factor out more code yet into individual header files nor does it change the logic of the server or the underlying logging library. All these changes are necessary but planned for later.

This improves testability at a later point in time.


src/go/cmd/http-relay-server/server/server.go
```diff
@@ -0,0 +1,530 @@
+package server
```
therjak, line 1: +license

therjak (approved):

PR #198, opened 2023-08-07
Switched to klog error messages

hauke76:
Improves parsing by fluentbit and reduces Stackdriver / Log Explorer noise.


ensonic:
FYI. Christian use a different log library in the token vendor already.

PR #203, opened 2023-08-14
Enable vertical autoscaling on GKE.

drigz:
This will be useful for pods that we can't easily scale horizontally and
are tolerant of downtime. Also, reformat the Terraform files to make the
pre-commit hook happy.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -103,16 +103,13 @@ function terraform_exec {
 }
 
 function terraform_cleanup {
-  # Terraform doesn't seem to remove node pool taints without recreating the
-  # node pool, whereas this approach causes less downtime.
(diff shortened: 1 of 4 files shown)


deploy.sh
```diff
@@ -103,16 +103,13 @@ function terraform_exec {
 }
 
 function terraform_cleanup {
-  # Terraform doesn't seem to remove node pool taints without recreating the
-  # node pool, whereas this approach causes less downtime.
-  pool=(base-pool "--cluster=cloud-robotics" "--zone=${GCP_ZONE}" "--project=${GCP_PROJECT_ID}")
```
vivekbagade, line 108: Do we not need these lines anymore?

drigz: Sorry, forgot to mention in the commit message. This is an old cleanup from when we first enabled Workload Identity last year.

vivekbagade (approved):

PR #206, opened 2023-08-18
Add option to configure Kubernetes QPS limit.

csieber:
We have seen that on the integration test project the current default limit of 5 is reached.


ensonic (approved):
thanks

drigz:
Thank you! If for some reason this doesn't work, we might need to configure `Burst` too, but from the doc on https://pkg.go.dev/k8s.io/client-go/util/flowcontrol I think this is sufficient and a good fit for the steady load of the token-vendor.

vivekbagade (approved):

PR #208, opened 2023-08-18
Reduce metric cardinality

drigz:
This fixes a typo in previous attempt to reduce k8s metric cardinality,
and adds a hack to reduce the broker metric cardinality. We should
probably avoid logging request paths in the broker entirely, but I'd
prefer we build our our backend metrics before losing this info.

I also had to rerun Gazelle to make the precommit hook happy and label a
flaky test.


vivekbagade (approved):
Thank you!

PR #209, opened 2023-08-21
Fix a chart-assignment-controller race

drigz:
The chart-assignment-controller expands and applies Helm charts in background goroutines by calling from controller.go into release.go. Then, it periodically checks the status ("phase") in a struct shared with the background goroutine and updates the ChartAssignment status.

When updating an existing chart, there is a race between the background goroutine (running `release.update()`), which changes the phase from Ready to LoadingChart, and the controller goroutine (calling `release.setStatus()` from `Reconciler.reconcile()`), which queries the phase and updates the status. I suspect that this race was introduced by a change that ironically described itself as "overall simplifying the controller", although the only repro I have so far is that it happens in <1% of our cloud integration tests...

While I'm here, I have protected access to the `generation` member, as the docs say that you can't treat non-atomic types as atomic.

Alternative considered: Tear out the async handling of Helm charts and do it all in the reconcile() method. I haven't figured out why the async logic was added in the first place. Removing it would mean that a single very large chart could block the handling of other charts, but I don't think that would be a problem. However, it would be a much bigger change than this.

b/292208405

### DIFF
--- a/src/go/pkg/controller/chartassignment/release.go
+++ b/src/go/pkg/controller/chartassignment/release.go
@@ -68,11 +68,11 @@ func newReleases(cfg *rest.Config, rec record.EventRecorder) (*releases, error) 
 
 // release is a cache object which acts as a proxy for Synk ResourceSets.
 type release struct {
-	name       string
-	synk       synk.Interface
-	recorder   record.EventRecorder
-	actorc     chan func()
-	generation int64 // last deployed generation.
+	name     string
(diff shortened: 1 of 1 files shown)


therjak (approved):

ensonic:
Thanks! It is unfortunate that SetStatus is doing so many phase checks.

ensonic (commented):
+1

PR #212, opened 2023-08-22
Add option to configure Kubernetes QPS limit (fixed burst parameter)

csieber:
We have seen that on the integration test project the current default limit of 5 is reached.

This second iteration of the PR fixes the missing burst parameter.

### DIFF
--- a/src/go/cmd/token-vendor/main.go
+++ b/src/go/cmd/token-vendor/main.go
@@ -65,6 +65,7 @@ var (
 		string(Kubernetes),
 		"Public key repository implementation to use. Options: "+strings.Join(keyStoreOpts, ","))
 	k8sQPS = flag.Int("k8s-qps", 25, "Limit of QPS to the Kubernetes API server.")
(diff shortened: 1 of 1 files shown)


drigz:
Thanks! How have you tested this?

csieber:
Tested with `./deploy.sh update robco-ensonic`. Seems to run fine now.

ensonic:
Do we need to ensure that Burst > QPS?

drigz:
> Do we need to ensure that Burst > QPS?

We'll crash with a semi-helpful error anyway. I've seen some code assume burst=2*qps, if we want to make this quicker for people to adjust.

drigz (approved):

drigz:
@csieber Do you plan to submit this or is there anything you're waiting for?

It's non-urgent (as the manual scaling seems to work OK) but if this is still the plan for mitigating the issue, it would be good to submit today, so we can deploy a daily release to giza-workcells this week and then to intrinsic-integration-tests + staging projects next week.

csieber:
@drigz I was under the assumption it is on auto-submit. Thanks for the notification!

PR #213, opened 2023-08-23
Escape double quotes in release notes body

faximan:
Attempt to fix current failures in e.g. https://github.com/googlecloudrobotics/core/actions/runs/5947260705/job/16129080081.

Tested by running the GitHub release creation part locally from this branch.

### DIFF
--- a/.github/ci/release_binary.sh
+++ b/.github/ci/release_binary.sh
@@ -47,7 +47,7 @@ output=$(curl -fsS \
   -H "Authorization: token $GITHUB_TOKEN" \
   https://api.github.com/repos/$REPO/releases/generate-notes \
   -d '{"tag_name":"'$RELEASE_NAME'","previous_tag_name":"'$PREVIOUS_RELEASE_NAME'"}')
-# Code newlines as literal \n and escape double quotes to make curl happy.
+# Code newlines as literal \n and escape double quotes to generate valid JSON.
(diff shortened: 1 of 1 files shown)


.github/ci/release_binary.sh
```diff
@@ -47,8 +47,8 @@ output=$(curl -fsS \
   -H "Authorization: token $GITHUB_TOKEN" \
   https://api.github.com/repos/$REPO/releases/generate-notes \
   -d '{"tag_name":"'$RELEASE_NAME'","previous_tag_name":"'$PREVIOUS_RELEASE_NAME'"}')
-# Code newlines as literal \n so thar curl understands that.
-BODY="$(jq -r '.body'   <<< $output | awk '{printf "%s\\n", $0}')"
+# Code newlines as literal \n and escape double quotes to make curl happy.
+BODY="$(jq -r '.body'   <<< $output | awk '{printf "%s\\n", $0}' | sed 's/"/\\"/g')"
```
drigz, line 51: I'd argue "to generate valid JSON when the commit message uses quotes" is more accurate, I think curl will happily send our invalid JSON to the GitHub API which will then return 400 Bad Request.

drigz (approved):

.github/ci/release_binary.sh
faximan: Good point - done. Thanks for the review Rodrigo!

Issue #207, opened 2023-08-18
http-relay-server/server:go_default_test is ~15% flaky


drigz:
```
bazel test //src/go/cmd/http-relay-server/server:go_default_test --runs_per_test=100
[...]
//src/go/cmd/http-relay-server/server:go_default_test                    FAILED in 18 out of 100 in 0.2s
```

errors look like:

```
server_test.go:74: Wrong encapsulated request; want id:"foo:52fdfc072182654f163f5f0f9a621d72" method:"GET" host:"example.com" url:"http://invalid/bar?a=b#c" header:{name:"X-Deadline" value:"now"} body:"body"; got 'id:"foo:52fdfc072182654f163f5f0f9a621d72" method:"GET" host:"example.com" url:"http://invalid/bar?a=b#c" header:{name:"Traceparent" value:"00-3b5c1eb13329b5e03091774bf5e6ef82-9d1e1bb088d48f1f-00"} body:"body"'
```

the diff here is `header:{name:"X-Deadline"...}` vs `header:{name:"Traceparent...}`. I'll mark the test as flaky so this won't block presubmits.

drigz:
```
git bisect start 07eddfa 14ae831
git bisect run bazel test //src/go/cmd/http-relay-server:go_default_test --runs_per_test=100
```
points to #172.

PR #216, opened 2023-08-29
Fixes issue #207

hauke76:
After adding tracing, it was a wrong assumption to assume that the X-Deadline was stored as the first entry in a header. Manually removing the traceparent turns out to be the correct and reliable solution.


ensonic (approved):
Thanks

PR #218, opened 2023-09-04
Fix handling of trailers in relay when chunking

ferstlf:
The relay client splits responses from the backend into multiple chunks based on MaxChunkSize and sends these chunks to the relay server. Currently, the relay client includes HTTP trailers in the last chunk sent - which makes sense - but the relay server code does not correctly handle the case with more than one chunk and always returns the trailers sent with the first chunk (ignoring the trailers in all other chunks). Thus when chunking responses we currently do not return any trailers. 

This PR fixes this issue in the relay server by handling trailers similar to the message body and collecting all trailers from all chunks before appending them to the HTTP response sent to the user client. We add an integration test and unit tests that cover chunked responses.

This fixes the issue that relaying gRPC requests with responses larger than MaxChunkSize leads to a `UNKNOWN` (`No status received`) error. This also removes an existing [TODO](https://github.com/googlecloudrobotics/core/blob/63b665ba7a0a9d9e46d0e387397ceade6db6659e/src/go/cmd/http-relay-server/server/server.go#L163) which has spotted exactly this issue before.

### DIFF
--- a/src/go/cmd/http-relay-server/server/server.go
+++ b/src/go/cmd/http-relay-server/server/server.go
@@ -129,40 +129,55 @@ func extractBackendNameAndPath(r *http.Request) (backendName string, path string
 	return
 }
 
+type responseChunk struct {
+    Body []byte
+	Trailers []*pb.HttpHeader
+}
+
 // responseFilter enforces that there's at least one HttpResponse in the out
-// channel and that the first response has a status code. From the reposnses it
-// extracts and return headers, trailers, status-code and body data.
-func responseFilter(backendCtx backendContext, in <-chan *pb.HttpResponse) ([]*pb.HttpHeader, []*pb.HttpHeader, int, <-chan []byte) {
-	body := make(chan []byte, 1)
+// channel and that the first response has a status code. From the responses it
+// extracts it synchronously returns headers and status-code and asynchronously
+// returns body and trailers via a channel.
+func responseFilter(backendCtx backendContext, in <-chan *pb.HttpResponse) ([]*pb.HttpHeader, int, <-chan *responseChunk) {
+	responseChunks := make(chan *responseChunk, 1)
 	firstMessage, more := <-in
+
 	if !more {
 		brokerResponses.WithLabelValues("client", "missing_message", backendCtx.ServerName, backendCtx.Path).Inc()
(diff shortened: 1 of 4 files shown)


src/go/cmd/http-relay-server/server/server.go
```diff
@@ -129,40 +129,55 @@ func extractBackendNameAndPath(r *http.Request) (backendName string, path string
 	return
 }
 
+type responseChunk struct {
+    Body []byte
```
ensonic, line 133: please indent

src/go/cmd/http-relay-server/server/server.go
```diff
@@ -129,40 +129,55 @@ func extractBackendNameAndPath(r *http.Request) (backendName string, path string
...
+}
+
 // responseFilter enforces that there's at least one HttpResponse in the out
-// channel and that the first response has a status code. From the reposnses it
-// extracts and return headers, trailers, status-code and body data.
-func responseFilter(backendCtx backendContext, in <-chan *pb.HttpResponse) ([]*pb.HttpHeader, []*pb.HttpHeader, int, <-chan []byte) {
-	body := make(chan []byte, 1)
+// channel and that the first response has a status code. From the responses it
+// extracts it synchronously returns headers and status-code and asynchronously
+// returns body and trailers via a channel.
```
ensonic, line 140: I think there is something wrong in the comment

src/go/cmd/http-relay-server/server/server.go
```diff
@@ -129,40 +129,55 @@ func extractBackendNameAndPath(r *http.Request) (backendName string, path string
...
 		// Flush remaining messages
 		for range in {
 		}
-		return nil, nil, http.StatusInternalServerError, body
+		return nil, http.StatusInternalServerError, responseChunks
 	}
-	body <- []byte(firstMessage.Body)
-	lastMessage := firstMessage
+
+    responseChunks <- &responseChunk{
```
ensonic, line 165: indent

src/go/cmd/http-relay-server/server/server.go
ferstlf: You were too quick with your review :) Now all files are auto-formatted.

ensonic (approved):
Thanks

PR #219, opened 2023-09-05
Update http relay dashboard

Ongy:
The total number of backends that have ever connected doesn't appear to be a good metric.

Getting a good guess at which clients are currently online is a more useful measure.


ensonic (approved):
Thanks

PR #220, opened 2023-09-07
fix leaking connections when user client closes connection

koonpeng:
When the user client closes the connection, it is not propagated to the backend. For simple "one shot" requests, there is no huge impact as the response would just be dropped, however for long running requests (chunked encoding or websockets), the backend doesn't know that the user client has closed the connection and it keeps sending new updates to the relay client, which also doesn't know and so keep forwarding that to the relay server.

The fix applied is:

* Add `StopRelayRequest` to the broker to "forget" a relaying request. This will cause the relay server to response with a permanent error the next time the relay client tries to send a response for that id. This will in turn cause the relay client to close the connection to the backend.
* There is a bug in the relay client when checking for backoff permanent errors. When the backoff operation encounters a permanent error, it actually unwraps it and return the underlying error, but the client is still checking for `backoff.Permanent`, resulting in the check to always fail and never handling it.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -657,6 +657,9 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 		)
 		if _, ok := err.(*RelayServerError); ok {
 			// A permanent error suggests the request should be aborted.
+			log.Printf("[%s] %s", *resp.Id, err)
+			log.Printf("[%s] Closing backend connection", *resp.Id)
+			hresp.Body.Close()
 			break
 		}
 	}
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -91,6 +91,18 @@ type ClientConfig struct {
 	ForceHttp2   bool
 }
 
+type RelayServerError struct {
+	msg string
+}
+
+func NewRelayServerError(msg string) error {
+	return &RelayServerError{msg}
+}
+
+func (e *RelayServerError) Error() string {
+	return e.msg
+}
+
 func DefaultClientConfig() ClientConfig {
 	return ClientConfig{
 		RemoteRequestTimeout:   60 * time.Second,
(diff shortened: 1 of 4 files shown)


src/go/cmd/http-relay-client/client/client.go
```diff
@@ -384,7 +396,7 @@ func (c *Client) postResponse(remote *http.Client, br *pb.HttpResponse) error {
 		return fmt.Errorf("couldn't read relay server's response body: %v", err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		err := fmt.Errorf("relay server responded %s: %s", http.StatusText(resp.StatusCode), body)
+		err := NewRelayServerError(fmt.Sprintf("relay server responded %s: %s", http.StatusText(resp.StatusCode), body))
 		if resp.StatusCode == http.StatusBadRequest {
 			// http-relay-server may have restarted during the request.
```
drigz, line 401: Please extend this comment to say "or the client cancelled the request."

src/go/cmd/http-relay-client/client/client.go
```diff
@@ -643,8 +655,11 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 				log.Printf("[%s] Failed to post response to relay: %v", *resp.Id, err)
 			},
 		)
-		if _, ok := err.(*backoff.PermanentError); ok {
+		if _, ok := err.(*RelayServerError); ok {
```
drigz, line 658: This is a change in behavior: Before, only HTTP 400 would terminate the loop, now anything other than HTTP 200 terminates the loop. For example, if the nginx ingress is overloaded and returns 500s for a while, this will now terminate the request. I think this is a good thing: There is no reason for us to drop a chunk from the middle of the response body, and it's better to terminate the request.

I think you could go even further: any error returned from RetryNotify indicates that we aren't sure we've posted the response, and that we should stop this loop rather than incorrectly continuing with the next response chunk. WDYT?

Please also delete the next comment which is no longer true (this could be a "transient" 5xx error that lasted too long). Maybe `// The relay server was unreachable for too long, so we dropped the chunk and should abort the request.`

drigz (commented):
Thank you, this is a really great change! And also a helpful commit message. I will add you to the organization so you can push to the `googlecloudrobotics/core` repo as otherwise the presubmits will fail with a permissions error.

src/go/cmd/http-relay-client/client/client.go
koonpeng: That sounds good, silently dropping chunks will cause clients to receive corrupted data, it breaks the normal assumption of http/tcp connections.

koonpeng:
Moved to #222

koonpeng:
Thanks for the comments, I updated the other PR with the suggestions here.

PR #222, opened 2023-09-13
fix leaking connections when user client closes connection

koonpeng:
Same as #220, but ported to this repo.

---

When the user client closes the connection, it is not propagated to the backend. For simple "one shot" requests, there is no huge impact as the response would just be dropped, however for long running requests (chunked encoding or websockets), the backend doesn't know that the user client has closed the connection and it keeps sending new updates to the relay client, which also doesn't know and so keep forwarding that to the relay server.

The fix applied is:

* Add `StopRelayRequest` to the broker to "forget" a relaying request. This will cause the relay server to response with a permanent error the next time the relay client tries to send a response for that id. This will in turn cause the relay client to close the connection to the backend.
* There is a bug in the relay client when checking for backoff permanent errors. When the backoff operation encounters a permanent error, it actually unwraps it and return the underlying error, but the client is still checking for `backoff.Permanent`, resulting in the check to always fail and never handling it.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -396,7 +396,7 @@ func (c *Client) postResponse(remote *http.Client, br *pb.HttpResponse) error {
 		return fmt.Errorf("couldn't read relay server's response body: %v", err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		err := NewRelayServerError(fmt.Sprintf("relay server responded %s: %s", http.StatusText(resp.StatusCode), body))
+		err := NewRelayServerError(fmt.Sprintf("relay server responded %s or the client cancelled the request: %s", http.StatusText(resp.StatusCode), body))
 		if resp.StatusCode == http.StatusBadRequest {
 			// http-relay-server may have restarted during the request.
 			return backoff.Permanent(err)
@@ -576,6 +576,7 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 	defer span.End()
 
 	resp, hresp, err := makeBackendRequest(ctx, local, req, id)
+	defer hresp.Body.Close()
 	if err != nil {
 		// Even if we couldn't handle the backend request, send an
 		// answer to the relay that signals the error.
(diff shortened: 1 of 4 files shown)


src/go/cmd/http-relay-client/client/client.go
```diff
@@ -643,8 +655,11 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 				log.Printf("[%s] Failed to post response to relay: %v", *resp.Id, err)
 			},
 		)
-		if _, ok := err.(*backoff.PermanentError); ok {
+		if _, ok := err.(*RelayServerError); ok {
 			// A permanent error suggests the request should be aborted.
+			log.Printf("[%s] %s", *resp.Id, err)
+			log.Printf("[%s] Closing backend connection", *resp.Id)
```
ensonic, line 661: Please only log a single line:
log.Printf("[%s] Relay error: %v, closing backend connection", *resp.Id, err)

drigz:
Thanks for moving the PR, the CI looks happy now. Will you also address the other comments from #220?

src/go/cmd/http-relay-client/client/client.go
koonpeng: updated in [301b04d](https://github.com/googlecloudrobotics/core/pull/222/commits/301b04d5a0c3449018d5207a18f145e812cf4d02)

src/go/cmd/http-relay-client/client/client.go
```diff
@@ -643,8 +656,11 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 				log.Printf("[%s] Failed to post response to relay: %v", *resp.Id, err)
 			},
 		)
-		if _, ok := err.(*backoff.PermanentError); ok {
-			// A permanent error suggests the request should be aborted.
+		// Any error suggests the request should be aborted.
+		// A missing chunk will cause clients to receive corrupted data, in most cases it is better
+		// to close the connection to avoid that.
+		if err != nil {
+			log.Printf("[%s] Closing backend connection (%s)", *resp.Id, err)
```
ensonic, line 663: You would use "%v" to log an error object.

ensonic (commented):
Thanks

src/go/cmd/http-relay-client/client/client.go
drigz: %w only works in fmt.Errorf, not in log.Printf. %v would be more consistent, I will fix it in a followup since I'd prefer not to block this PR any longer.

drigz (approved):

Issue #59, opened 2020-12-09
cr-syncer stops working on authorization errors


oliver-goetz:
When cr-syncer faces an *Unauthorized* error (I assume at Kubernetes API endpoint of the cloud cluster) it stops working completely and repeats this error message multiple times per second.

```E1209 15:26:45.385515       1 reflector.go:153] external/io_k8s_client_go/tools/cache/reflector.go:105: Failed to list *unstructured.Unstructured: Unauthorized```

In this state cr-syncer is not syncing any CR. After deleting the pod, cr-syncer starts working again.

This error usually occurs when you run the [setup_robot.sh](https://raw.githubusercontent.com/googlecloudrobotics/core/master/src/bootstrap/robot/setup_robot.sh) script again on a robot which is already setup. Maybe cr-syncer is running on an old (invalid) REST config for the remote Kubernetes API server when this happens.
As this error does not occur very often in my opinion it would be ok to simply crash cr-syncer on this error.

drigz:
I haven't dug into it, but I think "crash if we get unauthorized" is easier said than done, as https://github.com/kubernetes-sigs/controller-runtime is doing the requests for us, and that error is coming from the kubernetes/client-go source (reflector.go).

My guess is that setup-robot is creating a new robot key, and the metadata-server is restarting with that new key, but the cr syncer is not getting a new token from the metadata server. Do you know if the problem resolves itself after ~1 hour? Has the problem occurred without being caused by setup_robot.sh?

While fixing controller-runtime or client-go to either crash or fetch a new token when it gets "unauthorized" might be tricky, it could be easier to change setup-robot to avoid changing the identity if it has already been set up. I'm not sure why it does that right now, but it could just be that we didn't notice it causing this problem before. What do you think?

ensonic:
A stop gap fix could be to create some dependency in cr-syncer on the configmap with the secret. So that the cr-syncer yaml changes when setup_robot created a new key.
But I agree it would be nicer to avoid creating a new key if one already exists. If one needs a new key, it should be possible to simply delete the local secret.

drigz:
My theory above is probably wrong, I don't think tokens become "invalid" after the identity changes. I wasn't able to reproduce the problem by running setup_robot.sh and watching the cr-syncer logs, even if I restart the token-vendor to clear any cache there.

[...]

PR #223, opened 2023-09-14
Increase cr-syncer health check period

drigz:
On our integration test project, this makes up 90% of the requests from
the cr-syncer. I hope this will reduce the load on nginx.

b/299409898

### DIFF
--- a/src/app_charts/base/robot/cr-syncer.yaml
+++ b/src/app_charts/base/robot/cr-syncer.yaml
@@ -30,11 +30,11 @@ spec:
             port: 8080
           failureThreshold: 3
           initialDelaySeconds: 10
-          periodSeconds: 10
+          periodSeconds: 120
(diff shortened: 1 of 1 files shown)


mp4096 (approved):

ensonic (approved):

drigz:
> Why do the health checks go through the ingress?

The cr-syncer has an unconventional health-check that does a request from onprem to cloud to check the end-to-end auth/request flow, which we added to address an issue with invalid tokens: https://github.com/googlecloudrobotics/core/issues/59#issuecomment-865037601

The cr-syncer talks to GKE through the ingress because we set `--remote_server=www.endpoints.project.cloud.goog` - maybe we could avoid it by teaching it another way to determine the IP address of the remote GKE apiserver, but I'm not sure exactly how to do that, and maybe there's a good reason it uses the ingress.

PR #224, opened 2023-09-15
Use %v instead of %s in relay logs

drigz:
This doesn't change the formatting but is more consistent with our other
error format strings.

I also fixed the location of a string due to an unclear review comment I
made.


ensonic (approved):
thanks for the cleanup!

faximan:
I think this might have broken the postsubmit: https://github.com/googlecloudrobotics/core/actions/runs/6196097256

(happened to notice as I manually ran postsubmit for an unrelated change that failed with the same error)

PR #225, opened 2023-09-15
Update Bazel to 6.3.2

faximan:
Also minor improvements to the integration test image. The size is reduced from 4.17G to 4.05G.


mering (approved):
Thanks!

PR #227, opened 2023-09-16
Changed the http status code from 500 Internal Server Error to 504 Gateway Timeout after timing out from the backend

theaimboy:
Testedt by using curl through a relay-server to a slow backend, and seeing the relay-server responding with a "HTTP/2 504" error instead of 500 and "Timeout after 1m0s, indicating that the backend request took too long" in the body.


hauke76 (changes requested):
Can you please add a test which shows that we hit this branch after the timeout? Since now both the relay server and client are separate classes, it is possible to write such a test without the need to start multiple binaries for the client/server.

hauke76 (approved):

hauke76:
To unblock us in landing this change and the reverse proxy change, I unblocked the submission here despite the missing test. I will add an issue which tracks the test implementation.

hauke76:
Added https://github.com/googlecloudrobotics/core/issues/229 to track the implementation of the test.

drigz:
#230

PR #228, opened 2023-09-18
fix leaking connections 2nd try

koonpeng:
Fix end-to-end failure from #222.

The problem is mistakenly using `w.Write` at server.go:260, the response writer is no longer valid after the connection has been hijacked, we need to use `bufrw.Write` instead.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -91,6 +91,18 @@ type ClientConfig struct {
 	ForceHttp2   bool
 }
 
+type RelayServerError struct {
+	msg string
+}
+
+func NewRelayServerError(msg string) error {
+	return &RelayServerError{msg}
+}
+
+func (e *RelayServerError) Error() string {
+	return e.msg
+}
+
 func DefaultClientConfig() ClientConfig {
 	return ClientConfig{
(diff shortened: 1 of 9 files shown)


drigz (approved):
Thank you for the quick fix, and for adding the test to catch this quicker in future!

src/go/cmd/http-relay-client/client/client.go
```diff
@@ -564,6 +576,7 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 	defer span.End()
 
 	resp, hresp, err := makeBackendRequest(ctx, local, req, id)
+	defer hresp.Body.Close()
```
hauke76, line 579: Why are we closing the Body here? Does the comment below not apply anymore?

// hresp.Body is either closed from streamToBackend() or streamBytes()

Also, we are not trying segfaulting if err != nil since then Body is nil and we cannot close it.

hauke76 (commented):
Sorry, I toyed around with in-process unit tests and stumbled up this while doing so.

src/go/cmd/http-relay-client/client/client.go
ensonic: +1 I'd revert that line and see to have missed the change

PR #231, opened 2023-09-22
Fix a silly error in URL cleanup

drigz:
Previously it only removed exactly 20 characters of hex characters,
rather than 20+ characters, which severely limited its value. I promise
that this is my last attempt to save this duct-tape solution, and next
time I will remove request_path from the URL entirely.


ensonic (commented):
Thanks

PR #232, opened 2023-09-26
In-process relay tests

hauke76:
This is the first in-process relay test.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -668,18 +668,32 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 func (c *Client) localProxy(remote, local *http.Client) error {
 	// Read pending request from the relay-server.
 	relayURL := c.buildRelayURL()
-	req, err := c.getRequest(remote, relayURL)
-	if err != nil {
-		if errors.Is(err, ErrTimeout) {
-			return err
-		} else if errors.Is(err, ErrForbidden) {
-			log.Fatalf("failed to authenticate to cloud-api, restarting: %v", err)
-		} else if errors.Is(err, syscall.ECONNREFUSED) {
-			log.Fatalf("failed to connect to cloud-api, restarting: %v", err)
+
+	var req *pb.HttpRequest = nil
+	var err error = nil
+
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		req, err = c.getRequest(remote, relayURL)
+		if err != nil {
+			if errors.Is(err, ErrTimeout) {
+				return err
+			} else if errors.Is(err, ErrForbidden) {
+				log.Fatalf("failed to authenticate to cloud-api, restarting: %v", err)
+			} else if errors.Is(err, syscall.ECONNREFUSED) {
+				continue
+			} else {
+				return fmt.Errorf("failed to get request from relay: %v", err)
+			}
 		} else {
-			return fmt.Errorf("failed to get request from relay: %v", err)
+			break
 		}
 	}
+
+	if err != nil {
+		log.Fatalf("failed to connect to cloud-api, restarting: %v", err)
+	}
+
 	// Forward the request to the backend.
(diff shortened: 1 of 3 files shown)


drigz:
@koonpeng Can you review this? I haven't touched it so recently.

src/go/tests/relay/in_process_relay_test.go
```diff
@@ -0,0 +1,127 @@
...
+	once      sync.Once
+)
+
+func initRelay() {
+	once.Do(func() {
+		glog.Info("Running init.")
+
+		go func() {
+			relayServer := server.NewServer()
+			relayServer.Start(port, blockSize)
```
koonpeng, line 30: `nok8s_relay_test.go` let's the OS auto assign a free port, it then scans stdout to learn which port is used. I think we can improve on it here by using port `0` and getting the actual port used after the server is started.

The catch is that `relayServer.port` will not contain the actual port used in this case, we need to update it after we start listening. https://github.com/googlecloudrobotics/core/blob/41c3e73120d1a431e699c798f36eece275e54952/src/go/cmd/http-relay-server/server/server.go#L534

src/go/tests/relay/in_process_relay_test.go
```diff
@@ -0,0 +1,127 @@
...
+	return serveFunctionWithTimeout(f, 10*time.Second)
+}
+
+func serveFunctionWithTimeout(
+	f func(w http.ResponseWriter, r *http.Request),
+	handlerTimeout time.Duration) *http.Server {
+	httpHandlerFunc := http.HandlerFunc(f)
+
+	srv := &http.Server{
+		Addr:         "127.0.0.1:8083",
```
koonpeng, line 75: Same as above, we should be able to let the OS assign a free port and check which port is used.

src/go/tests/relay/in_process_relay_test.go
```diff
@@ -0,0 +1,127 @@
...
+	if !bytes.Equal(observedResponse, expectedResponse) {
+		t.Errorf("Received wrong response.\n\tExpected: %s\n\tObserved: %s", expectedResponse, observedResponse)
+	}
+}
+
+func TestHttpTimeout(t *testing.T) {
+	initRelay()
+
+	httpServer := serveFunctionWithTimeout(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(2 * time.Second)
```
koonpeng, line 114: Can I check what is the behavior we want to test? iiuc, this sleep will cause the server to return `503` error, but the server is our test backend server. If we want to test if the relay propagates this error, can we write the response directly, without this sleep slowing down the test?

src/go/cmd/http-relay-client/client/client.go
```diff
@@ -669,18 +669,32 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
...
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		req, err = c.getRequest(remote, relayURL)
+		if err != nil {
+			if errors.Is(err, ErrTimeout) {
+				return err
+			} else if errors.Is(err, ErrForbidden) {
+				log.Fatalf("failed to authenticate to cloud-api, restarting: %v", err)
+			} else if errors.Is(err, syscall.ECONNREFUSED) {
+				continue
```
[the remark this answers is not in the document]
hauke76: I added a log message.

hauke76 (commented):
Thanks for your review. I addressed all comments.

src/go/tests/relay/in_process_relay_test.go
koonpeng: I tried changing this to

```go
httpServer := serveFunction(func(w http.ResponseWriter, r *http.Request) {
  w.WriteHeader(http.StatusServiceUnavailable)
})
```

and it appears to work. Would this be more preferrable?

hauke76: Hm, I think the timeout test is fine. In the end it also works. We could add this as another tests, since this tests if any response in the header is forwarded.

koonpeng (approved):

PR #233, opened 2023-09-27
Fix trying to close nil connection

koonpeng:
I just learned that the behavior of multiple Close is undefined so we probably should avoid calling it multiple times https://github.com/golang/go/blob/5152be3959d4aa273932c12da971d14e7f84405d/src/io/io.go#L103-L109.

I think there are 3 scenarios here, either backend closes the connection, the remote (relay-server) closes, or the remote returns an error.

1. If the backend closes, either streamToBackend or streamBytes can close hresp.Body, but streamToBackend only runs for websocket connections.
2. If the remote closes, streamToBackend is the only one that knows about it.
3. If the remote returns an error, streamToBackend is the only one that knows about it.

It makes sense to close it in streamToBackend, but streamToBackend only runs for websocket connections. It works if we also close in streamBytes, but that leads to a racey condition and causes us to close multiple times. It's probably better if we close it in the else branch of line ~588 (if *resp.StatusCode == http.StatusSwitchingProtocols {), this way we ensure that we only ever call Close() once.


drigz (approved):

src/go/cmd/http-relay-client/client/client.go
```diff
@@ -600,6 +597,10 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 		}
 		// Stream stdin from remote to backend
 		go c.streamToBackend(remote, id, bodyWriter)
+	} else {
+		// `streamToBackend` will close `hresp.Body` but it is only ran on websocket connections.
```
hauke76, line 601: nit: s/ran/called

hauke76 (approved):

drigz:
Submitting this because @Ongy  saw it causing segfaults on startup of the kubernetes-relay-client.

PR #237, opened 2023-10-05
Revert "Allow robot-service@ to read from GCR explicitly. (#234)"

drigz:
This reverts commit 4f2065c322c12b77372f526c1f32c789ff9ec2fe. It breaks
on projects with "eu.artifacts..." buckets like the "navtest" CI
project.


faximan (approved):
Thanks!

PR #239, opened 2023-10-16
Use --fail-with-body

drigz:
This should provide more debugging information on error.

This requires an update to 22.04 to get a newer curl. This might break after merging if  I didn't do the update in the right places.

### DIFF
--- a/.github/ci/presubmit.sh
+++ b/.github/ci/presubmit.sh
@@ -5,6 +5,9 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# DO NOT SUBMIT
+curl --version
+
 # shellcheck source=ci/common.sh
 source "${DIR}/common.sh"
 
--- a/.github/ci/presubmit.sh
+++ b/.github/ci/presubmit.sh
@@ -5,9 +5,6 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# DO NOT SUBMIT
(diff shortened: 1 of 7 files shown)


.github/ci/presubmit.sh
```diff
@@ -5,6 +5,9 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+# DO NOT SUBMIT
+curl --version
+
```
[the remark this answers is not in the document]
drigz: Thanks for the review, although feel free to wait until I add someone as reviewer - this is in here to test the presubmit and will be removed when the change is ready. I'll also try to add WIP in future to avoid wasted review cycles.

.github/workflows/release.yml
```diff
@@ -21,7 +21,7 @@ env:
 
 jobs:
   create_release:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
faximan, line 24: Maybe let's also bump presubmit.yml for consistency? https://github.com/search?q=repo%3Agooglecloudrobotics%2Fcore%20ubuntu-20.04&type=code

I'd also feel comfortable with `ubuntu-latest` if you think that'd reduce work over time.

faximan (approved):
Nice! TIL!

.github/workflows/release.yml
drigz: Thanks, went with 22.04 for now, maybe it's nicer not to be surprised when latest changes.

PR #240, opened 2023-10-20
Rate-limit token requests on error

drigz:
If the token-vendor encounters an error (such as being rate-limited by
the k8s apiserver), the metadata-server will retry as fast as it gets
requests from on-prem clients, which can be quite fast, as they don't
expect a round-trip to the cloud but just a local request.

This change applies exponential backoff with a max delay of around 2
minutes. I tested by turning off the token-vendor in robco-rodrigoq,
waiting for the metadata-server's token to expire, waiting another hour
or so, then turning it back on, and observing that it got a token within
a few minutes.

Alternative considered: Instead of using the TokenSource interface, I
could have created a goroutine that periodically refreshes the token and
does normal retries on failure, but this approach feels more composable?
(although the metadata-server log output is very spammy when the
token-vendor is unavailable, because it prints the error for every local
token request)

Fixes b/298948451.

### DIFF
--- a/src/go/cmd/metadata-server/metadata.go
+++ b/src/go/cmd/metadata-server/metadata.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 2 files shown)


ensonic (approved):
Thanks!

src/go/cmd/metadata-server/metadata.go
```diff
@@ -42,6 +43,59 @@ const (
...
 
+// rateLimitTokenSource is a TokenSource that applies exponential backoff on
+// errors, returning the previous error if called again too soon. It doesn't
+// rate-limit on success, as it assumes it wraps a reuseTokenSource.
+type rateLimitTokenSource struct {
+	wrapped oauth2.TokenSource
+
+	mu       sync.Mutex // guards err and next
+	err      error
+	exponent int
```
therjak, line 54: would the code be simpler if it would state the delay and maxDelay instead of the exponent and maxExponent on initialDelay?

drigz: I don't think so - we need to bound the value of exponent to avoid overflows (my first version counted "errors" rather than "exponent" but if "errors" got too big then `initialDelay << errors` would wrap around) - or are you imagining a different structure?

src/go/cmd/metadata-server/metadata.go
```diff
@@ -42,6 +43,59 @@ const (
 	getPodByIPWait = 500 * time.Millisecond
 )
 
+// rateLimitTokenSource is a TokenSource that applies exponential backoff on
+// errors, returning the previous error if called again too soon. It doesn't
+// rate-limit on success, as it assumes it wraps a reuseTokenSource.
```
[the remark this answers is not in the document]
drigz: Done.

src/go/cmd/metadata-server/metadata.go
therjak: your max delay is 102,4sec. It still feels save to double it a few times without overflow. And the upper bound would be even more explicit as it wouldn't be initial * 2^exponent but a written constant.
I was considering something similar to:

type rateLimetTokenSource struct {
delay time.Duration
...
}
func newRateLimitTokenSource(ts oauth2.TokenSource) *rateLimitTokenSource {
	return &rateLimitTokenSource{wrapped: ts, delay: 100 * time.Millisecond}
}
func (s *rateLimitTokenSource) resetBackoff() {
	s.delay = 100 * time.Millisecond
	s.next = time.Time{}
}
func (s *rateLimitTokenSource) updateBackoff(err error) {
	if s.delay <= 51200 * time.Millisecond {
		s.delay *= 2
	}
	s.next = timeNow().Add(s.delay)
	s.err = err
}

drigz: Thanks for explaining, that is indeed simpler.

therjak (approved):

PR #243, opened 2023-11-03
Allow robot-service@ to read from GCR explicitly.

drigz:
This is a step towards removing the blanket objectAdmin ACL.
(b/299475780) It's a second attempt after the rollback in #237.

As far as I can tell, robot-service@ does not need to read other files
from CRC buckets (config.sh, crc_version.txt, etc) in the general case,
although integration tests might need more privilege.


faximan (approved):
Thanks!

PR #247, opened 2023-11-10
Response with 503 error if relay client is not connected

koonpeng:
b/309909333

Currently when the relay server receives a user client request before the relay client is connected, it waits for the relay client until a timeout and returns a 500 error. This change it such that if the request is to a backend that the relay server has not yet seen before, it immediately returns with a 503 error, it also adds a `X-CLOUDROBOTICS-HTTP-RELAY` header containing the request id.


src/go/cmd/http-relay-server/server/broker.go
```diff
@@ -91,6 +91,12 @@ type pendingResponse struct {
 	requestPath string
 }
 
+type RelayClientUnavailableError struct{}
+
+func (*RelayClientUnavailableError) Error() string {
+	return "relay client not available"
```
drigz, line 97: Can you keep the error string from before? It makes it easier to debug when people have the wrong server name:

`Cannot reach the client %q. Check that it's turned on, set up, and connected to the internet.`

Before we had the suffix `(timeout waiting for relay client to accept request)` maybe here something like `(unknown client)` is better.

src/go/cmd/http-relay-server/server/server.go
```diff
@@ -377,6 +377,11 @@ func (s *Server) userClientRequest(w http.ResponseWriter, r *http.Request) {
 	// from the relay-client.
 	backendRespChan, err := s.relayRequest(ctx, *backendCtx, backendReq)
 	if err != nil {
+		if _, ok := err.(*RelayClientUnavailableError); ok {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			w.Header().Set("X-CLOUDROBOTICS-HTTP-RELAY", backendCtx.Id)
+			return
```
drigz, line 383: Please also return an error body here because people can see this error in their browser (and I think also in inctl output etc). `err.Error()` should be fine with the suggestion above.

drigz (changes requested):
Thanks! FYI @Ongy

drigz:
Also FYI @sbarfurth not sure how often this comes up, but this changes the error status code for a never-before-seen onprem cluster from 500 to 503. It will also affect the case where the onprem cluster has been disconnected for several days (ie since before the last server process restart).

sbarfurth:
I think it shouldn't make a difference whether the error is a 503 or a 500. Thanks for the heads up.

src/go/cmd/http-relay-server/server/broker.go
koonpeng: done

src/go/cmd/http-relay-server/server/server.go
koonpeng: done

drigz (approved):

PR #249, opened 2023-11-10
update used go compiler to 1.21

therjak:
also fixes broken gazelle in go/tests/relay


drigz (approved):

faximan (approved):
Thank you @therjak for cc'ing me! The sysroot changes might be what we were missing to bump this in the internal repo too - I'll have a look!

PR #250, opened 2023-11-14
Reduce permissions on gke-node service account.

ensonic:
* monitoring.viewer is already covered by monitoring.metricWriter
* stackdriver.resourceMetadata.writer is unused

Tested on robco-ensonic gcp project.


drigz:
The stackdriver/monitoring roles are recommended here: https://cloud.google.com/kubernetes-engine/docs/troubleshooting/troubleshoot-dashboards#write_permissions but not sure if they're actually necessary or if that's out of date.

drigz (approved):

PR #251, opened 2023-11-16
Hide integration tests from Buildkite CI

drigz:
The Bazel team hosts a CI project on Buildkite, which expects to be able
to run all our tests with recent versions of Bazel. We like this because
the Bazel team file issues on our repo to tell us when a future Bazel
release will break our build, but to keep this, we need to make sure our
project is always green on their CI. I don't know why, but we just got a
report that our integration tests are red there due to missing
dependencies (kubectl, docker, credentials): #4380.

We can work around this by using the "manual" tag to exclude our tests
from target expansion, and then manually listing them in our own CI.

I will need to manually check that the tests are still running in
presubmit and postsubmit.

### DIFF
--- a/.github/ci/integration_test.sh
+++ b/.github/ci/integration_test.sh
@@ -140,7 +140,7 @@ bazel_ci test \
   --test_output=streamed \
   --test_tag_filters="external" \
   --strategy=TestRunner=standalone \
-  //...
+  //... //src/go/tests:go_default_test //src/go/tests:relay_test
 
 # If this is running on main (ie, not a manual run) then update the `latest`
 # binary.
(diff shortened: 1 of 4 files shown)


ensonic (approved):
Thanks.

drigz:
Confirmed:

```
# postsubmit before
Step #1: //src/go/tests:go_default_test                                           PASSED in 1.2s
Step #1: //src/go/tests:relay_test                                                PASSED in 16.5s
# postsubmit after
Step #1: //src/go/tests:go_default_test                                           PASSED in 1.4s
Step #1: //src/go/tests:relay_test                                                PASSED in 8.4s
```

PR #260, opened 2023-11-29
http-relay-client: Use HTTP/2 PING to recover from network issues

drigz:
This should fix issues we encountered when (re)configuring the network
on the nodes. b/313612734


drigz:
nginx did not catch on fire when deploying this to our test cluster, so I'm optimistic this won't overload anything.
image

drigz:
FYI @koonpeng

hauke76 (approved):

ensonic (commented):
Thanks

PR #271, opened 2023-12-04
migrating to log/slog

therjak:


therjak:
this is as far as I got. From the previous log setup I assume we should change the ilog/handler to print time in RFC3339Nano but this should also require an update to the fluentbit filter.

drigz:
> we should change the ilog/handler to print time in RFC3339Nano but this should also require an update to the fluentbit filter

It doesn't seem so important. The previous format wasn't well parsed either, and now we're more likely to use traces for sub-second debugging rather than text logs.

I checked the effect of the changes to the app-rollout-controller logging on GKE: The new format is not parsed (GKE's fluent-bit configuration only handles glog and json logs, not our format). I tried switching to JSON logs, which seems to handle the nanosecond timestamp and level correctly, but because slog.NewJSONHandler uses `msg` instead of `message`, they don't look nice in the Cloud Logging UI (you see the full JSON object rather than just the message).

We have a choice between this: 
image

or this: 
image

or [deploying a custom fluent-bit](https://cloud.google.com/knowledge/kb/how-to-deploy-a-custom-fluent-bit-on-google-kubernetes-engine-000004839)...

drigz:
I filed b/315475658 to see if there's a chance to get `msg` treated the same as `message` in JSON payloads - but it might be worth it to enable JSON logging for things that already use log.Lmicroseconds to get more precise handling in Cloud Logging.

Tangentially, I also found a command to see already-used keys that might be useful: `grep -rho 'slog.String("[^"]*"' | sort -u`.

drigz:
@therjak Here's the progress I made: https://github.com/googlecloudrobotics/core/commits/rodrigoq/slog/

- replaced all lines matching `\blog\.`
- switch cloud services to use JSON logging (so GKE's fluent-bit can parse the severity and timestamp, seems like the lesser of two evils)
- switched to stderr for consistency with before (made this change when I saw the nok8s_relay_test failing, as it parses the relay port from stderr...)

Let me know what you think, I wonder if we should send this as one big PR or split it up into a few smaller changes, so Christian can review the token-vendor changes for example.

therjak:
all but gcr-credential-refresher are migrated.  gcr-credential-refresher is expected to get removed

PR #272, opened 2023-12-05
Add retry for TestIamPermissions on context deadline exceeded.

csieber:
This change adds retries with static retry interval to the TestIamPermissions calls. Exponential retries should not be necessary here with the small number of retries.

### DIFF
--- a/src/go/cmd/token-vendor/oauth/verifier.go
+++ b/src/go/cmd/token-vendor/oauth/verifier.go
@@ -52,13 +52,8 @@ func (v *TokenVerifier) Verify(ctx context.Context, token Token, sa string) erro
 			iamActAs, resource)
 	}
 	// query IAM if not found in cache
-	preq := iam.TestIamPermissionsRequest{
-		Permissions: []string{iamActAs},
-	}
-	pcall := v.s.Projects.ServiceAccounts.TestIamPermissions(resource, &preq)
-	pcall.Header().Set("Authorization", "Bearer "+string(token))
 	ts := time.Now()
-	resp, err := pcall.Context(ctx).Do()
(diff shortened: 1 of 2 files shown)


src/go/cmd/token-vendor/oauth/verifier.go
```diff
@@ -71,6 +66,44 @@ func (v *TokenVerifier) Verify(ctx context.Context, token Token, sa string) erro
...
+	pcall := s.Projects.ServiceAccounts.TestIamPermissions(resource, &preq)
+	pcall.Header().Set("Authorization", "Bearer "+string(token))
+
+	for i := 0; i <= retries; i++ {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+		ctx, cancel := context.WithTimeout(ctx, timeout)
```
drigz, line 87: WDYT about doubling the timeout on retry? If TestIamPermissions starts taking consistently 2.1s (or whatever) then this will be a major regression. The p99 latency here is 2s: https://g3doc.corp.google.com/apphosting/g3doc/slos/admin_api.md#blocking-reads so allowing more for the retries feels conservative.

csieber: I increased the timeout to 5s. If we further increase the cumulative time the request takes (max currently 5+1+5+1+5=17s) I think it is better if the client retires instead.

drigz: I was thinking of 2+1+4+1+8+1=17 but allowing 5s seems enough if the p99 SLO is 2s. IIRC we've never seen high p50 latency, only high p99 latency.

drigz (approved):

PR #273, opened 2023-12-06
Delete obsolete CR_SYNCER_RBAC flag

drigz:
This has been in use for a while. However, CR_SYNCER_RBAC is not
consistently true across all our projects - leading to excessive
robot-service@ privileges in some case. Deleting this simplifies the
code and improves security.

After this is applied, robot-service@ will be limited to "Kubernetes
Engine Cluster Viewer" in all other projects, which allows it to list
GKE clusters (via IAM) and other Robot CRs and ChartAssignments (via
RBAC) but not much more. This role is not to be confused with
"Kubernetes Engine Viewer" which would allow reading all resources in
the GKE clusters, rather than just listing clusters and reading select
resources.

### DIFF
--- a/src/bootstrap/cloud/terraform/service-account.tf
+++ b/src/bootstrap/cloud/terraform/service-account.tf
@@ -71,8 +71,7 @@ resource "google_project_iam_member" "robot-service-roles" {
 resource "google_project_iam_member" "robot-service-kubernetes" {
   project = data.google_project.project.project_id
 
-  # TODO(rodrigoq): migrate to RBAC and remove roles/container.developer
-  role = var.cr_syncer_rbac == "true" ? "roles/container.clusterViewer" : "roles/container.developer"
+  role = "roles/container.clusterViewer"
 
   member = "serviceAccount:${google_service_account.robot-service.email}"
 }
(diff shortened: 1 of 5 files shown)


src/bootstrap/cloud/terraform/service-account.tf
```diff
@@ -71,8 +71,7 @@ resource "google_project_iam_member" "robot-service-roles" {
 resource "google_project_iam_member" "robot-service-kubernetes" {
   project = data.google_project.project.project_id
 
-  # TODO(rodrigoq): migrate to RBAC and remove roles/container.developer
-  role = var.cr_syncer_rbac == "true" ? "roles/container.clusterViewer" : "roles/container.developer"
+  role = "roles/container.clusterViewer"
 
   member = "serviceAccount:${google_service_account.robot-service.email}"
 }
```
ensonic, line 77: can you fold this into the robot-service-roles?

drigz: Done, I'm worried that Terraform won't cope with the rename so I've added a line to deploy.sh to make sure it's maintained.

drigz:
PTAL

ensonic (commented):
I think we don't need the gcloud in deploy.sh since this is the same terraform. But happy to play safe and remove this later.

PR #275, opened 2023-12-12
Remove unused buildcleaner.

drigz:
I can't find any instructions/docs that refer to this. This also lets us remove an unused dependency.
I stumbled across it when looking at the outstanding work for Carsten's
log->slog migration.


third_party/go_repositories.bzl
faximan, line 1: Is this change intentional? Maybe add a note about it to the PR description.

faximan (approved):
Thanks!

third_party/go_repositories.bzl
drigz: Yep, it's unused now.

PR #279, opened 2024-01-06
Update .bazelrc

limdor:
Make sure that no regressions are introduced until the flag is flipped


ensonic (approved):
Tested locally via `gh pr checkout 279` and `bazel test //...`

PR #281, opened 2024-01-10
Make letsencrypt-prod a ClusterIssuer

### LINKED ISSUE #276: Simplify workload ingress TLS configuration for specific domains
This project is currently configured to serve all of its cloud workloads from the same domain. Configuring a workload to be served from a different domain requires an extra set of steps that are quite hacky and can be simplified with a small change.

## Context

The configuration deployed by this repository introduces a configuration for the `ingress-nginx-controller` that defaults the TLS certificates of all cluster ingresses to the `cloud-robotics` certificate.

```yaml
$ kubectl edit ingress-controller

[...]
metadata:
  name: nginx-ingress-controller
[...]
    spec:
      containers:
      - args:
        [...]
        # `default/tls` is a secret with the contents of the TLS certificate for the `cloud-robotics` Kubernetes Certificate resource
        - --default-ssl-certificate=default/tls 
[...]
```

Because this is configured as the default certificate, ingresses in this repository that enable TLS get the `cloud-robotics` certificate assigned automatically. This works well because all workloads are expected to be served from the same domain.

I had a workload that had to be accessed from a different domain. Enabling SSL/TLS access for this workload required creating a new `Certificate` Kubernetes resource for the target domain and assigning it to the workload ingress through the `tls` field.

The issue I ran into is that the `letsencrypt-prod` certificate issuer is of type `Issuer` and not `ClusterIssuer`. Because of its type, it can only emit certificates in the same Kubernetes namespace the issuer is defined, in this case `default`. 

The workload ingress I was attempting to configure was in an application namespace. Because the Kubernetes Ingress resource only supports referencing TLS certificates by secret name, but secrets cannot be shared between Kubernetes namespaces by design, it is not possible to configure the ingress to reference certificates in other namespaces.

I did find a workaround in [this comment](https://github.com/kubernetes/ingress-nginx/issues/2170#issuecomment-1127969684). 

Because of the way `ingress-nginx-controller` works, all certificates required by cluster ingresses are loaded into memory if required by an ingress, independent of if the ingress actually contains paths or not. I was able to load the newly created certificate into the `ingress-nginx-controller` instance by creating a dummy ingress in the `default` namespace that did nothing other than reference the certificate secret and associate it with the target domain (without any path rules), and then created a second ingress in the application namespace with path rules and that only mentioned the target domain in the `tls` field, without mentioning the secret. This works but is a bit hacky.

## Solution Suggestion

Changing the `letsencrypt-prod` issuer from type `Issuer` to type `ClusterIssuer` will allow applications to define certificates inside their namespaces, and avoid having to create two ingress configurations, one for loading and one for its actual usage.

I am opening this as an issue as I don't know the security ramifications of this and to discuss other potential options we can implement.

marcoshuck:
Fix #276

### DIFF
--- a/src/app_charts/base/cloud/cert-manager-issuers.yaml
+++ b/src/app_charts/base/cloud/cert-manager-issuers.yaml
@@ -15,6 +15,9 @@ spec:
     secretName: cluster-authority
 {{ if eq .Values.certificate_provider "lets-encrypt" }}
 ---
+# While an Issuer may satisfy our current needs within the default namespace,
+# anticipating future growth and potential deployments in additional namespaces,
+# adopting a ClusterIssuer offers a more scalable and versatile solution.
 apiVersion: cert-manager.io/v1
 kind: ClusterIssuer
 metadata:
--- a/src/app_charts/base/cloud/cert-manager-issuers.yaml
+++ b/src/app_charts/base/cloud/cert-manager-issuers.yaml
@@ -16,7 +16,7 @@ spec:
 {{ if eq .Values.certificate_provider "lets-encrypt" }}
 ---
 apiVersion: cert-manager.io/v1
-kind: Issuer
+kind: ClusterIssuer
 metadata:
   name: letsencrypt-prod
 spec:


AlejoAsd:
Has this been deployed to a cluster to test? I would like to verify that a certificate can be created in a namespace that is not `default`.

marcoshuck:
> Has this been deployed to a cluster to test? I would like to verify that a certificate can be created in a namespace that is not `default`.

src/app_charts/base/cloud/cert-manager-issuers.yaml
```diff
@@ -16,7 +16,7 @@ spec:
 {{ if eq .Values.certificate_provider "lets-encrypt" }}
 ---
```
ensonic, line 17: Please add a comment such as:
# We need kind to be ClusterIssuer to be able to issue verts for multiple namespaces. If all we need would be the default namespace Issuer would be sufficient.

src/app_charts/base/cloud/cert-manager-issuers.yaml
```diff
@@ -15,8 +15,11 @@ spec:
     secretName: cluster-authority
 {{ if eq .Values.certificate_provider "lets-encrypt" }}
 ---
+# While an Issuer may satisfy our current needs within the default namespace,
+# anticipating future growth and potential deployments in additional namespaces,
+# adopting a ClusterIssuer offers a more scalable and versatile solution.
```
AlejoAsd, line 20: Why does it offer a more scalable solution? Instead of talking about future needs, maybe try specifying what a `ClusterIssuer` allows us to do that a regular `Issuer` doesn't.

ensonic (approved):

src/app_charts/base/cloud/cert-manager-issuers.yaml
ensonic: I suggested a different wording above, but marcos did not wanted to use it and I am so busy today that I merged to unblock you. Feel free to send a new PR with a better working.

AlejoAsd: Thanks Stefan. This will allow us to test and eventually get another PR in if necessary.

marcoshuck: > Why does it offer a more scalable solution?

It offers a more scalable solution in the sense that more applications can be deployed using the same `ClusterIssuer`. Let's add more clarifications in a subsequent PR, I was trying to keeping it as abstract as possible.

PR #282, opened 2024-01-11
Log the error when we don't retry again.

ensonic:
This will more consistently log when a chart fails.


methylDragon (approved):
This will go a long way to ensuring we have some visibility into failures.

methylDragon:
Might be a good idea to keep track of retry count and explicitly fail if it crosses a threshold though?

vivekbagade:
> Might be a good idea to keep track of retry count and explicitly fail if it crosses a threshold though?

We should only count errors and fatalf in the case of client-go failures and not other failures. We don't want a single bad chartassignment bringing down the CAC

ensonic:
I'll merge this then. If we keep logging an error due to infinite retries, it will be annoying, but at least visible.

PR #264, opened 2023-11-30
Bump k8s.io/client-go past v0.27.0 including transitively

methylDragon:
When I ran `go mod graph | grep client-go` after the merge of:
- https://github.com/googlecloudrobotics/core/pull/263

I saw that our direct dependencies were still using `v0.25.X` of `k8s.io/client-go` (transitively.)

e.g.
```
github.com/prometheus/prometheus@v0.39.1 k8s.io/client-go@v0.25.1
k8s.io/apiextensions-apiserver@v0.25.3 k8s.io/client-go@v0.25.3
k8s.io/cli-runtime@v0.25.3 k8s.io/client-go@v0.25.3
k8s.io/component-base@v0.25.3 k8s.io/client-go@v0.25.3
sigs.k8s.io/controller-runtime@v0.13.1 k8s.io/client-go@v0.25.0
```

We need a more recent version of client-go all throughout the stack to support: https://github.com/googlecloudrobotics/core/issues/262

So this is my attempt to upgrade all dependencies which use client-go transitively.
I'm not too strong in golang dependency management, so if I shouldn't have done it this way, please let me know!!

Is there another smarter way to force all transitive declarations of a particular dependency to be of a minimum version?

I ran:
```
go get -u sigs.k8s.io/controller-runtime
go get -u github.com/prometheus/prometheus
go get -u k8s.io/apiextensions-apiserver
go get -u k8s.io/cli-runtime
go mod tidy
```

Then `go mod graph | grep client-go` gives something which makes it seem like `client-go` got upgraded all throughout the dependency tree:
```
github.com/googlecloudrobotics/core/src k8s.io/client-go@v0.28.4
github.com/prometheus/prometheus@v0.48.0 k8s.io/client-go@v0.28.2
k8s.io/apiextensions-apiserver@v0.28.4 k8s.io/client-go@v0.28.4
k8s.io/cli-runtime@v0.28.4 k8s.io/client-go@v0.28.4
k8s.io/client-go@v0.28.4 github.com/evanphx/json-patch@v4.12.0+incompatible
k8s.io/client-go@v0.28.4 github.com/gogo/protobuf@v1.3.2
k8s.io/client-go@v0.28.4 github.com/golang/groupcache@v0.0.0-20210331224755-41bb18bfe9da
k8s.io/client-go@v0.28.4 github.com/golang/protobuf@v1.5.3
k8s.io/client-go@v0.28.4 github.com/google/gnostic-models@v0.6.8
[...]

PR #283, opened 2024-01-11
Downgrade "empty response" error in discovery client to info log for metrics

methylDragon:
This follows the precedent set in client-go's [memcache](https://github.com/kubernetes/client-go/commit/0bc91705fa50a531b8a1ee9dca3ef063ca71bb0c).

The discovery client's `ServerGroupsAndResources()` call sometimes throws up empty response errors if a metrics server (namely GCP's custom metric server) neglects to populate its resource list. This is not considered an error.

The [bump of client-go's version](https://github.com/googlecloudrobotics/core/pull/264) fixed `memcache.go`, but we also need to account for it in our code.

Tested with a test deploy, the CAC works as expected.

---

**Note**: Logging the downgraded error generates quite a few spurious logs on startup (in fact, I think it generates one log PER chart update call, so each time a chart is updated, we'll get a printout...) Might be scary to a user, so we might consider just not printing, throttling the print, or explicitly writing "(NOT AN ERROR)" in the log. What do y'all think?

I am quite for just ignoring the error. So much so I'm going to push that change now.

image

### DIFF
--- a/src/go/pkg/synk/synk.go
+++ b/src/go/pkg/synk/synk.go
@@ -481,9 +481,15 @@ func (s *Synk) populateNamespaces(
 	_, span := trace.StartSpan(ctx, "Discover server resources")
 	_, list, err := s.discovery.ServerGroupsAndResources()
 	span.End()
+
+	// Don't treat "empty response" from external metrics servers as an error.
+	// Follows precedent from: https://github.com/kubernetes/client-go/commit/0bc91705fa50a531b8a1ee9dca3ef063ca71bb0c
 	if err != nil {
-		return errors.Wrap(err, "discover server resources")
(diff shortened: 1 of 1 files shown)


vivekbagade (approved):

vivekbagade (commented):
Make the check and exception specific to external.metrics.k8s.io, in any other case, log the API group and return error

methylDragon:
> Make the check and exception specific to external.metrics.k8s.io, in any other case, log the API group and return error

https://github.com/googlecloudrobotics/core/compare/c1419c3378b9676987af1ccdf3090052e93859fa..eefe73b183a3d67e792ef52eb72e57468fd6b43f

vivekbagade (approved):

PR #286, opened 2024-01-15
Remove invalid part of certificate spec.

ensonic:
This fixes a warnign on the cert validation. I wend back in the git history and can't figre why we have these lines here. These belong to the issuer config and are not valid fro Certificates, see
https://cert-manager.io/docs/usage/certificate/


drigz (approved):
I haven't touched this code so this is not a very well informed approval: Your rationale in the description makes sense but it's possible Marcos or Alejo would be better informed reviewers.

ensonic:
> I haven't touched this code so this is not a very well informed approval: Your rationale in the description makes sense but it's possible Marcos or Alejo would be better informed reviewers.
The snippet was there before @AlejoAsd extended the file. But I am happy to wait for their confirmation.

PR #295, opened 2024-01-16
Simplify rbe_config

faximan:
We realized the only thing used from the rbe_config directory is a single `platform` rule in `bazel/rbe_config/config/BUILD`. This is because the full hermetic toolchain is specified through WORKSPACE, we do not use the https://github.com/bazelbuild/bazel-toolchains auto generated one.

Move this rule to the parent BUILD and rename it to be more explicit. Delete the rest of the directory.

Also adds `.bazelversion` to make this repo bazelisk compatible.


faximan:
> I have no idea what I'm reviewing here so if you're uncertain that this is right please find someone who knows about RBE to review it! But it looks fine on the face of it and if it works it works.

Thanks for taking a look! This corresponds to e.g. cl/598746015 which is being reviewed by other team members, so I'll wait with merging this PR until we have aligned fully internally. That being said, I'm pretty confident in this.

PR #308, opened 2024-01-22
removed logrus dependency

therjak:
the dependency is already removed in all packages. this removes the last references.


ensonic (approved):
Thanks!

PR #309, opened 2024-01-22
pkg/controller/chartassignment: switch to log/slog

therjak:


src/go/pkg/controller/chartassignment/release.go
```diff
@@ -194,14 +196,14 @@ func (r *release) setGeneration(generation int64) {
...
-		log.Printf("chart failed in phase %v: %v", r.status.phase, err)
+		slog.Warn("chart failed", slog.Any("phase", r.status.phase), ilog.Err(err))
 		// We only update the phase for non-retriable errors. This mitigates a
 		// race condition between ensureUpdated, which sets phase=Updating when
 		// retrying, and setStatus, which reads either the old phase or Updating
 		// and copies it to the chartassignment status.
 		r.status.phase = apps.ChartAssignmentPhaseFailed
 	} else {
-		log.Printf("chart failed in phase %v (retrying): %v", r.status.phase, err)
+		slog.Warn("chart failed (retrying)", slog.Any("phase", r.status.phase), ilog.Err(err))
```
ensonic, line 206: why Any instead of String?

ensonic (approved):
Thanks

src/go/pkg/controller/chartassignment/release.go
therjak: phase is not of type string. my assumption is slog.Any works similar to '%v'. as the underlying type of phase is string, we can cast it to string to pass it via slog.String but initially I did not check the actual type of phase.

PR #311, opened 2024-01-23
Adjust node-pool limit.

ensonic:
Allow to scale down to 1 if there is not a lot of load (might especially help on -ar clusters and dev setups). Bump max size on base-pool to 16 as we have hit that limit in some setups.


ensonic:
I've tested this on robco-ensonic. It is not yet scaling down to 1 node though.

mp4096 (approved):

PR #318, opened 2024-01-31
Extend the backup script to also backup the pubkeys.

ensonic:
This is mostly FYI. I don't like the hacks it requires. If I run this against g-w I get a 0.5 mb yaml and there are a lot of pubkeys:
```
egrep "^kind:" /tmp/crc-backup.yaml | sort | uniq -c
    706 kind: ConfigMap
     78 kind: Robot
```
Some of them are stale, but token-vendor can't do the cleanup, because the robot-cr is not necessarily available in the cloud.

Also the filtering is sub-par. WDYT if I change token-vendor to label them (and backfill the label via script).

### DIFF
--- a/scripts/backup_robots.sh
+++ b/scripts/backup_robots.sh
@@ -34,4 +34,9 @@ function kc {
 
 kc get robots -o yaml | \
   yq 2>/dev/null -ry '.items[] | del(.metadata.annotations["kubectl.kubernetes.io/last-applied-configuration"],.metadata.creationTimestamp,.metadata.generation,.metadata.managedFields,.metadata.resourceVersion,.metadata.selfLink,.metadata.uid,.status)' -
-
+echo "---"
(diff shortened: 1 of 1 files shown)


drigz:
> WDYT if I change token-vendor to label them (and backfill the label via script).

Sounds like an easy win.

> Some of them are stale, but token-vendor can't do the cleanup, because the robot-cr is not necessarily available in the cloud.

I guess we could add our own ownerReferences-like annotation for the ones that do have a Robot CR so they can be cleaned up, and for the dev-* ones we would need some kind of expiry logic, eg bumping a timestamp when you run setup_local_cluster.sh and deleting after 90d. Or if we can migrate everyone to VM pools we could delete all that don't correspond to a robot CR. Either way it would need some change to the token-vendor to propagate the timestamp/owner info through...

drigz (approved):

PR #329, opened 2024-02-09
Increase retention size of robot prometheus to 512MB

vivekbagade:
Current 128 MB is not enough even for the WAL files


ensonic:
@drigz is is okay to change the default here, or should it be parametrized and modified through the app-rollout on a case by case basis?

drigz:
It's fine to change it here: Not sure if it will trigger a recreation of the pod or just reconfigure Prometheus live, but either is fine, especially since most on-prem Prometheus pods have full volumes and need deleting anyway. However, I don't understand the description `128 MB is not enough even for the WAL files` vs the comment:

```
    # `retentionSize`.  This doesn't include the WAL (~3h of data) and the
    # "head chunk".
```

Won't increasing this number just make the problem worse? I'd expect we need to reduce the metric cardinality and/or increase `sizeLimit: 512Mi` to allow for the very large WAL that we have with the current cardinality.

vivekbagade:
Hey @drigz. This PR goes with the doc I sent to @ensonic. Since this is a public PR, I skipped the doc link here.

Basically the retention limit applies to WAL+persisted chunks+mmapped chunks size, but when the limit is hit only the persisted chunks are cleared which can lead to a case where WAL keeps building up (we saw a buildup upto 400 Mi) and consumes all of the 128 Mi. Leaving enough space for WAL should keep prometheus more healthy.

I updated the CL description to add this info

src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -62,7 +62,7 @@ prometheus:
     # `retentionSize`.  This doesn't include the WAL (~3h of data) and the
```
drigz, line 62: Please fix the comment.

drigz (approved):
Thanks! I think I was primarily mislead by the incorrect comment (which I think I wrote, or at least never realized was wrong, sorry!)

PR #332, opened 2024-02-13
Improve cloud prometheus scalability

vivekbagade:
Remove high cardinality metrics from kubelet, kube-apiserver, kube state exporter, and node exporter. We also corrected the regex usage for existing dropped metrics


vivekbagade:
I tested this on one of our robots to confirm the dropped metrics are in fact being dropped. I'll perform a scalability test for upto 200 robots after releasing this test

src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -140,6 +145,11 @@ kubelet:
 nodeExporter:
   enabled: true
   serviceMonitor:
+    metricRelabelings:
+    # Drop high cardinality metrics from node exporter.
+    - action: drop
+      regex: "node_cpu_seconds_total"
+      sourceLabels: [ __name__ ]
```
ensonic, line 152: Is this also labeled with e.g. 'core' to cause high cardinality? Is this used in the performance grafana dashboards? (The one we print after inctl app start).

ensonic (approved):

src/app_charts/prometheus/prometheus-robot.values.yaml
drigz: The dashboard is based on `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate`. Based on the name, I think that's not based on this metric so will hopefully survive, but I'm not sure how to check which rules/dashboards will be broken by this change.

On this metric specifically, I was surprised it's so big, and I think your intuition is right. It's 11k metrics on giza-workcells, ranging from 48 to 944 per cluster, where xfa-box-32 is the worst. `cpu` seems to be the worst label (0..109) but `instance` and `pod` are slightly noisy (random IPs / node exporter pod IDs, but correlated with the cluster/node). It would be tempting to sum without `cpu`, `instance`, and `pod` if we want to keep this metric - but I don't know how easy that is to do in the relabelling config - and I don't want to block this change since Prometheus federation doesn't work right now.

vivekbagade:
We'll undoubtedly break a few dashboards with this change, but hopefully we'll get federation working properly again. We can try to add back some useful metrics one by one dropping their high cardinality labels.

PR #335, opened 2024-02-14
Allow configuration of Prometheus ingress auth.

csieber:


ensonic (approved):
We can submit as this, but I don't think we would need to support different auth urls for prom and graana.

PR #337, opened 2024-02-16
Update kube-prometheus-stack to 56.6.2

vivekbagade:
Also, make minor changes to cluster.tf as 1.24 goes obsolete.

This was deployed and tested on private GCP project


ensonic (approved):
Tested on robco-ensonic too. Works like a charm! Thanks

PR #338, opened 2024-02-19
chart-assignment-controller: use the same resource builder as synk

therjak:
see b/207673353
there was a difference between manifest parsing in synk and in the chart-assignment-controller. the handling of 'kind: List' was causing an infinite resource creation by the controller but was handled correctly by synk. with this change manifest parsing is done by cli-runtime/pkg/resource as is already the case for synk.

### DIFF
--- a/src/go/pkg/controller/chartassignment/release.go
+++ b/src/go/pkg/controller/chartassignment/release.go
@@ -34,7 +34,7 @@ import (
 	"go.opencensus.io/trace"
 	core "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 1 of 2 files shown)


src/go/pkg/controller/chartassignment/release.go
```diff
@@ -393,6 +393,7 @@ func newHTTPGetter(url, certFile, keyFile, caFile string) (getter.Getter, error)
 }
 
 func decodeManifests(manifests map[string]string) (res []*unstructured.Unstructured, err error) {
+	var resources []*unstructured.Unstructured
```
ensonic, line 396: Please use the output variable 'res', fine for me if you want to rename it

ensonic (approved):
Thanks! Would it be hard to add a test?

ensonic (approved):

drigz (approved):

PR #341, opened 2024-02-21
Reduce the size of the broker metrics.

ensonic:
Make the path an optional attribute on the metric. We added that to see if there is a performance difference in the backend path. We're not using it in the dashbards for the relay though.
Put it behind a flag, so that we can enable it to debug performance.

### DIFF
--- a/src/go/cmd/http-relay-server/main.go
+++ b/src/go/cmd/http-relay-server/main.go
@@ -90,10 +90,11 @@ import (
 
 var (
 	port      = flag.Int("port", 80, "Port number to listen on")
-	blockSize = flag.Int("block_size", 10*1024,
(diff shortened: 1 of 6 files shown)


ensonic:
@drigz do you think this is overkill and I should just rip out the attribute?

drigz:
> @drigz do you think this is overkill and I should just rip out the attribute?

Sorry, seems like we raced on this: https://github.com/googlecloudrobotics/core/pull/340

I would indeed lean towards removing the metric. If it's not available on the systems with high scale then it's not very useful as that's really where we would need it. But I only every used it as a replacement for our missing metrics - have you found request_path useful for something else?

ensonic:
I'll start over after https://github.com/googlecloudrobotics/core/pull/340 has been merged.

Issue #336, opened 2024-02-16
ERROR: Failed to load Starlark extension '@@llvm_toolchain//:toolchains.bzl'. in CI at Bazel@HEAD



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3660#018db012-ef59-4353-8ce7-f1c8864e4398

 Platform: Ubuntu

 Logs: 
```
ERROR: Failed to load Starlark extension '@@llvm_toolchain//:toolchains.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @@llvm_toolchain
 - @@llvm_toolchain_llvm
This could either mean you have to add the '@@llvm_toolchain_llvm' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
(04:02:39) INFO: Repository llvm_toolchain instantiated at:
  /var/lib/buildkite-agent/builds/bk-docker-rp17/bazel-org-repo-root/core/WORKSPACE:46:15: in <toplevel>
  /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ee5637b7b4bf3d4f87dda1c28f44d63f/external/toolchains_llvm/toolchain/rules.bzl:60:14: in llvm_toolchain
Repository rule toolchain defined at:
  /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ee5637b7b4bf3d4f87dda1c28f44d63f/external/toolchains_llvm/toolchain/rules.bzl:33:28: in <toplevel>
(04:02:39) ERROR: Error computing the main repository mapping: cycles detected during computation of main repo mapping
``` 
 Culprit: 

 Steps:
  ``` 
 git clone https://github.com/googlecloudrobotics/core
git reset 265e6b5a878255e7c5b5a389e26ec018a9338921  --hard
export USE_BAZEL_VERSION=76b9ee7317f5537ca74a1313a4af4bba783a46e2
bazel build //... 
 ``` 
 CC Greenteam @mai93

ensonic:
I can't repro it. fro me it builds: https://source.cloud.google.com/results/invocations/5664ead2-7627-45bb-81ad-7e8ac74482e1

ensonic:
[...]

Issue #331, opened 2024-02-13
Error occurred during the fetch of repository 'com_grail_bazel_toolchain':



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3654#018da047-c239-48d4-a7bc-dbc0d0b077c7

Platform: Ubuntu

Logs:

```
ERROR: An error occurred during the fetch of repository 'com_grail_bazel_toolchain':
   Traceback (most recent call last):
    File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/495bf8cb7a2d4eef2c464609d7ac8070/external/bazel_tools/tools/build_defs/repo/http.bzl", line 139, column 45, in _http_archive_impl
        download_info = ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://github.com/grailbio/bazel-toolchain/archive/0.8.1.tar.gz] to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/495bf8cb7a2d4eef2c464609d7ac8070/external/com_grail_bazel_toolchain/temp57152805545568202/0.8.1.tar.gz: Checksum was 3bb45f480e3eb198f39fc97e91df2c2fc0beaabbea620ba3034ac505786a1813 but wanted 751bbe30bcaa462aef792b18bbd16c401af42fc937c42ad0ae463f099dc04ea2
```

Steps:

```
 git clone -v https://github.com/googlecloudrobotics/core.git
 git reset 41d638b90b5c3aa6def69b76a13c20a2328f9c53 --hard
export USE_BAZEL_VERSION=3bdeaea0282a3031566d3a17756fc9d59ec7f131
bazel --nosystem_rc --nohome_rc version 
```

Note :  [https://github.com/grailbio/bazel-toolchain](https://github.com/grailbio/bazel-toolchain) was just transferred to [https://github.com/bazel-contrib/toolchains_llvm](https://github.com/bazel-contrib/toolchains_llvm)

And causing issue with checksum update.

CC Greenteam @mai93

ensonic:
Thanks!

This is from here: https://github.com/googlecloudrobotics/core/blob/main/WORKSPACE#L10

We'll update this in the next days.

PR #343, opened 2024-02-26
Update remaining urls for new toolchain rules.

ensonic:
I've forgot to change the download url and the doc url in the comment.

See #331, #336


faximan (approved):
Thanks!
Seems https://github.com/grailbio/bazel-toolchain redirects to https://github.com/bazel-contrib/toolchains_llvm, so I don't think this will have any visible effect, is that correct?

ensonic:
Yes, but eventually that redirect will go away and I want to make this cleaner.

Issue #346, opened 2024-03-07
install_k8s_on_on_robot.sh is broken


faximan:
It tries to install a single node kubeadm k8s 1.21 cluster, but the old apt repo is no longer valid

```
The repository 'http://apt.kubernetes.io/ kubernetes-xenial Release' no longer has a Release file.
```

Using the approach described in https://forum.linuxfoundation.org/discussion/864693/the-repository-http-apt-kubernetes-io-kubernetes-xenial-release-does-not-have-a-release-file works with k8s 1.24 and newer, but the script needs furhter modifications, see https://github.com/googlecloudrobotics/core/pull/345 for work in progress.

Another option is to retire this completely or move to e.g. k3s.

faximan:
This script is used to set up `robot-sim`, which is a VM used by `//src/go/tests:relay_test` in the CI.

We also have `//src/go/tests:go_default_test`, but I cannot actually tell which VM it is using [here](https://github.com/googlecloudrobotics/core/blob/main/src/go/tests/k8s_integration_test.go#L222-L232)

PR #347, opened 2024-03-07
Fix postsubmit by skipping install_k8s_on_robot.sh

faximan:
... which is broken: https://github.com/googlecloudrobotics/core/issues/346

This will only work until robot-sim needs to be re-installed, so it is a ticking time bomb.

Also fixed Bazel invocations, this still runs (only) the two tests previously explicitly mentioned.

### DIFF
--- a/.github/ci/integration_test.sh
+++ b/.github/ci/integration_test.sh
@@ -82,10 +82,13 @@ if [[ -z "$SIM_HOST" ]] ; then
   exit 1
 fi
 
-DEPLOY_FILES="src/bootstrap/robot/setup_robot.sh \
-  src/bootstrap/robot/install_k8s_on_robot.sh \
-  ./bazel-out/../../../external/kubernetes_helm/helm"
-init_robot_sim ${SIM_HOST} "${DEPLOY_FILES}"
+# BUG(https://github.com/googlecloudrobotics/core/issues/346)
+# install_k8s_on_robot is currently broken. The rest of the test will work
+# for as long as our old VM is functional.
+# DEPLOY_FILES="src/bootstrap/robot/setup_robot.sh \
+#   src/bootstrap/robot/install_k8s_on_robot.sh \
+#   ./bazel-out/../../../external/kubernetes_helm/helm"
+# init_robot_sim ${SIM_HOST} "${DEPLOY_FILES}"
 
 # Setup new robot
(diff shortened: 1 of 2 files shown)


faximan:
Green run here: https://pantheon.corp.google.com/cloud-build/builds;region=europe-west1/119d69fe-9ce0-43db-9f50-a33e9646cf1e?project=robco-integration-test

src/go/tests/BUILD.bazel
```diff
@@ -28,10 +28,7 @@ sh_test(
     data = [
         "@kubernetes_helm//:helm",
     ],
-    tags = [
-        "external",
-        "manual",
-    ],
+    tags = ["external"],
```
drigz, line 31: I think we still need manual so that `bazel test //...` works on CloudTop without the extra setup of the sim VM (test_env stuff).

.github/ci/integration_test.sh
```diff
@@ -140,7 +143,7 @@ bazel_ci test \
   --test_output=streamed \
   --test_tag_filters="external" \
   --strategy=TestRunner=standalone \
-  //... //src/go/tests:go_default_test //src/go/tests:relay_test
+  //...
```
drigz, line 146: Did you check this still runs k8s_integration_test.go? I thought that `manual` means it's not matched by //...
https://github.com/googlecloudrobotics/core/blob/e5d771c1e7060198f2fb6d1ea5a7f5740efe181f/src/go/tests/BUILD.bazel#L33

drigz (approved):

.github/ci/integration_test.sh
drigz: It does, so my understand is faulty, not the script:

```
//src/go/tests:go_default_test                                           PASSED in 11.3s
//src/go/tests:relay_test                                                PASSED in 10.8s
```

src/go/tests/BUILD.bazel
faximan: This still works because https://github.com/googlecloudrobotics/core/blob/main/.bazelrc#L42.

If we add "manual" that means we need to explicitly list the tests in the test script, which I'm now avoid. But let me know if you prefer to be explicit.

.github/ci/integration_test.sh
faximan: See above.

Issue #319, opened 2024-02-01
install_k8s_on_robot.sh Failed: Error: Failed to install docker E: Version '5:20.10.6~3-0~ubuntu-focal' for 'docker-ce' was not found 


tiesus:
When install the cluster on the robot via `install_k8s_on_robot.sh` i get the following error due to a missmatch of the docker versions:

`E:_ Version '5:20.10.6~3-0~ubuntu-focal' for 'docker-ce' was not found
ERROR: Failed to install docker-ce=5:20.10.6~3-0~ubuntu-focal. Try:
    sudo apt update && sudo apt install docker-ce=5:20.10.6~3-0~ubuntu-focal `

My docker version on the device is **24.0.5.** Is it really necessary to stick to the docker version, you require in the script?

drigz:
It appears the script is quite obsolete. Even if we lift the docker pin it will try to install k8s 1.21, which is likely to fail. A few years ago it served as a useful example to get started with a local k8s cluster, but now it's more trouble than it's worth. We ought to remove the obsolete script and the docs that point at it but I'm not quite sure how to replace them.

You could try other lightweight k8s distros like k3s or k0s as they will probably work better on your device - sorry that I can't point you at more up-to-date instructions!

drigz:
Note for readers: GitHub generates the text "closed as completed" but this is actually "closed as obsolete" because we deleted the script and changed the docs to say "set up your own cluster". That said, I'm not optimistic that someone could follow the docs as we haven't tested this end to end and there might be some config incompatibilities between the user cluster and our installation that we haven't documented how to resolve, in particular I'm thinking about the --pod-cidr flag: https://github.com/googlecloudrobotics/core/blob/8a86fa6bbfb8afc547fe4b2b28ceab9b71b0ad84/src/go/cmd/setup-robot/main.go#L60

PR #348, opened 2024-03-08
Make integration test use GKE robot

faximan:
This fixes https://github.com/googlecloudrobotics/core/issues/346 and fixes https://github.com/googlecloudrobotics/core/issues/319. It removes `install_k8s_on_robot.sh` and `setup-robot-sim.sh`, which are now deprecated.

We can run both the relay_test and the go_default_test against a "GKE robot" instead of one running in a VM. This removes a lot of complexity from the testing infrastructure. The downside is that we lose some test coverage of the onprem metadata server, but that seems like a price we have to pay.

Another advantage is that with this change, `integration_test.sh` can run locally and not only in the Cloud Build environment, which is useful for debugging.

Green run: https://github.com/googlecloudrobotics/core/actions/runs/8233748802/job/22513900247

Follow-ups: Remove robot-sim TF. Remove TF around custom cloud build pools, which were only needed for ssh to GCE.


drigz (approved):

drigz:
Looks like "this should be easy" was not very accurate... thank you for figuring this out!!

drigz:
We should also close #319 which is obsolete after the deletion of this script.

PR #352, opened 2024-03-18
Update owner group of CI projects

faximan:
I manually deleted `roles/owner` for robco@google.com last week, as I did not find the group in terraform. I did not realize it is applied through this mechanism.

Now the tests fail to re-apply with our friend [ORG_EXTERNAL_GROUP_OWNERS_DISALLOWED](https://github.com/googlecloudrobotics/core/actions/runs/8321070075/job/22766751996). However, this group has ~30 people, many/most of which do not have a relation to the project anymore. So switch `CLOUD_ROBOTICS_SHARED_OWNER_GROUP` to the second owners group that we already use, which seems more appropriate for the task.


ensonic (approved):
Thanks.
Let me also prune the robco@gogole.com email list a bit.

PR #353, opened 2024-04-02
cr-syncer: reduce log level of the 2 most noisy messages

therjak:
reduce the log level of
'Got Event' and 'Copied status to upstream' from Info to Debug. They are responsible for nearly all log entries.


drigz (approved):

drigz:
Great payoff for the migration to a real logging library!

PR #355, opened 2024-04-05
Increase retention size of robot prometheus to 1024

vivekbagade:
The last increase got us from 12 hours of retention to 20 hours of retention. This increase will extend us to 24 hours and give us some headroom.


src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -61,7 +61,7 @@ prometheus:
     # Historical data is limited to the lower of `retention` and
     # `retentionSize`.
     retention: "1d"
-    retentionSize: 512MB
+    retentionSize: 1024MB
```
drigz, line 64: Unless I'm missing something, increasing retentionSize without increasing sizeLimit below will make the problem worse, not better. I'd suggest either reducing this to "384MB" (to have some headroom to before getting disk space errors at 512MiB) and reducing retention to "3h" - or if it's unavoidable, setting sizeLimit to 1024MB and setting retentionSize to something a bit less than that.

src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -60,7 +60,7 @@ prometheus:
     logLevel: warn
     # Historical data is limited to the lower of `retention` and
     # `retentionSize`.
-    retention: "1d"
+    retention: "3h"
     retentionSize: 512MB
```
drigz, line 64: I'm thinking of sending a followup to reduce retentionSize slightly to add some margin from sizeLimit, because IIUC if set correctly, retentionSize should protect us from disk full if the `WAL + head chunk < 512Mi`, yet in the reported case it did not. Setting retention will fix it as long as `3h history + WAL + head chunk < 512Mi`, which covers the current situation but seems slightly less safe than the first condition.

Let me know if you see a reason not to do that!

drigz (approved):

src/app_charts/prometheus/prometheus-robot.values.yaml
drigz: Created https://github.com/googlecloudrobotics/core/pull/356.

drigz (approved):

PR #356, opened 2024-04-10
Create headroom between retentionSize and sizeLimit

drigz:
Also update the comments. I hope this will further reduce the risk of running out of disk space.

### DIFF
--- a/src/app_charts/prometheus/prometheus-robot.values.yaml
+++ b/src/app_charts/prometheus/prometheus-robot.values.yaml
@@ -59,18 +59,19 @@ prometheus:
 
     logLevel: warn
     # Historical data is limited to the lower of `retention` and
-    # `retentionSize`.
+    # `retentionSize`. WAL+maxChunkSize also counted toward retentionSize,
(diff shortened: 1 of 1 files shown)


src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -59,18 +59,19 @@ prometheus:
...
     logLevel: warn
     # Historical data is limited to the lower of `retention` and
-    # `retentionSize`.
-    retention: "1d"
-    retentionSize: 512MB
+    # `retentionSize`. WAL+maxChunkSize also counted toward retentionSize,
+    # but Prometheus puts no limit on WAL size, so it's up to us to make
+    # sure we don't have too many metrics and the WAL doesn't grow too big.
+    retention: "3h"
+    # If you increase retentionSize, increase sizeLimit below as well. We don't
```
ensonic, line 66: ... but keep in mind that this is RAM.

ensonic (approved):
Thanks

PR #359, opened 2024-04-18
Add log when health check fails

koonpeng:
This simply logs when a health check fails to make it easier to debug when kubernetes tries to restart it.

### DIFF
--- a/src/go/cmd/http-relay-server/server/server.go
+++ b/src/go/cmd/http-relay-server/server/server.go
@@ -199,6 +199,11 @@ func newBackendContext(r *http.Request) (*backendContext, error) {
 }
 
 func (s *Server) health(w http.ResponseWriter, r *http.Request) {
+	timer := time.AfterFunc(8*time.Second, func() {
+		slog.Warn("While checking server health, unable to acquire lock after 8 seconds")
+	})
+	defer timer.Stop()
+
 	if err := s.b.Healthy(); err != nil {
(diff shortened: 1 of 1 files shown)


src/go/cmd/http-relay-server/server/server.go
```diff
@@ -200,6 +200,7 @@ func newBackendContext(r *http.Request) (*backendContext, error) {
 
 func (s *Server) health(w http.ResponseWriter, r *http.Request) {
 	if err := s.b.Healthy(); err != nil {
+		slog.Error("Health check failed", ilog.Err(err))
```
drigz, line 203: Since the health check just locks a mutex, I'd be worried that this will never execute: Either the server is deadlocked and Healthy() never returns, or it's healthy and it returns nil.

I think we'd need to detect the deadlock by observing the case where either Lock() or Healthy() hangs for longer than 10 seconds and print the log statement in that case. Not sure if there's a race with Kubernetes killing the pod so maybe 8-9s is safer.

koonpeng: https://github.com/googlecloudrobotics/core/pull/359/commits/5ba9554156264e79672f39d09c81185ed5697da8 Added a 8 secs timer to print a warning.

drigz: Really clean solution, I was imagining something with channels and selects and so on. Thank you!!!

drigz (approved):

PR #377, opened 2024-06-13
Allow bypassing manual-run protection

drigz:
This means that when a release fails, you can push a fix to main and rerun to create the daily release.


ensonic (approved):
Good idea.

drigz:
Seems to work as MANUAL_RUN is false here:

image

PR #378, opened 2024-06-13
Use GitHub's concurrency control

drigz:
This avoids needing to manually delete lock files after timeouts.


ensonic (approved):
Nice

drigz:
Tested with manual runs.

PR #386, opened 2024-06-24
Only consider scheduled postsubmit runs

drigz:
Without this parameter, the API will also return commits from PRs, which
may not have been reviewed and could be malicious.

This has the downside that if you're trying to fix the release workflow
you need to change the schedules or be limited to one attempted fix per
day, but the alternatives (see below) are also not ideal.

For defense in depth, we also check against the head_repository id, which I tested by temporarily changing `event=schedule` to `actor=totoro642`, [in which case the action fails](https://github.com/googlecloudrobotics/core/actions/runs/9662992790/job/26654091527) with:

```
Unexpected head repository ID: 817231418 - check postsubmit.yml configuration
```

Alternatives considered:

- Changing postsubmit to run on `push` and then having the release workflow filter with `event=push` instead. This would make it easier to manually trigger a release from a recently submitted change, but doesn't address the (more common) issue that the Postsubmit fails due to some unrelated cloud-side breakage, and must be retriggered after intervening on the GKE cluster to get a green run.
- Changing postsubmit to directly trigger the release action on the current commit: This would let us remove the `git clone` pattern that makes the attack possible, but at the cost of allowing actions to trigger other actions, which feels like it could open another class of attacks (such as triggering a release from older, insecure code).
- Changing release to trigger on tag creation rather than a schedule. This would mean that only people with permissions to create tags can create releases, which is probably more standard on GitHub , but diverges from the "autopush" model we're using so would need discussion with other team members.

http://b/348316770

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,8 +33,12 @@ jobs:
           set -euo pipefail
           output=$(curl --fail-with-body -sS \
             -H "Accept: application/vnd.github+json" \
-            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
             "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&event=schedule&status=success")
+          repo_id=$(jq -r '.workflow_runs[0].head_repository.id' <<< $output)
+          if [[ "${repo_id}" != "${{ github.repository_id }}" ]] ; then
+            echo >&2 "Unexpected head repository ID: ${repo_id} - check postsubmit.yml configuration"
+            exit 1
+          fi
           sha=$(jq -r '.workflow_runs[0].head_sha' <<< $output)
           echo "latest_green=$sha" >> $GITHUB_OUTPUT
       - name: Checkout repository
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
           output=$(curl --fail-with-body -sS \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&status=success")
+            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&event=schedule&status=success")
           sha=$(jq -r '.workflow_runs[0].head_sha' <<< $output)
           echo "latest_green=$sha" >> $GITHUB_OUTPUT
       - name: Checkout repository


.github/workflows/release.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           output=$(curl --fail-with-body -sS \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&status=success")
+            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&event=schedule&status=success")
```
mering, line 37: You could also filter for `event=workflow_dispatch` but I am not sure if the query parameter allows specifying multiple events combined by a logical `OR`.

mering (commented):
> I [...] wasn't immediately sure how to do that check on the result of the API call (is there a better way than git log main | grep $green_sha?)

In order to check if a commit is part of a branch, you can use the following snippet:
```sh
if [ -z "$(git branch -a "origin/main" --contains "${green_sha}")" ]; then
  echo "Commit ${green_sha} is not part of the main branch. Aborting." >&2
  exit 1
fi
```

I also suggest to consider the suggestion from @nuclearsandwich-ai to check additional fields on the response via `jq`.
For example:
```sh
$ curl 'https://api.github.com/repos/googlecloudrobotics/core/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&status=success' | jq '.workflow_runs.[].head_repository.full_name'
"googlecloudrobotics/core"
```
You should probably even filter by `head_repository.id` which although less readable might stay constant even when the repository is moved or renamed.

.github/workflows/release.yml
drigz: `event=schedule,workflow_dispatch` seems not to work so we'd have to do it with jq instead. The result of limiting to event=schedule is to require waiting a day to release new changes but we can live with this.

.github/workflows/release.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           output=$(curl --fail-with-body -sS \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
```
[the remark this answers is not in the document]
drigz: Thanks, seems good in principle to not send tokens where they're not needed.

.github/workflows/release.yml
```diff
@@ -30,11 +30,15 @@ jobs:
...
+          set -euxo pipefail
           output=$(curl --fail-with-body -sS \
             -H "Accept: application/vnd.github+json" \
-            -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&status=success")
+            "https://api.github.com/repos/$REPO/actions/workflows/postsubmit.yml/runs?per_page=1&branch=main&event=schedule&status=success")
+          repo_id=$(jq -r '.workflow_runs[0].head_repository.id' <<< $output)
+          if [[ "${repo_id}" != "${{ github.repository_id }}" ]] ; then
+            echo >&2 "Unexpected head repository ID: ${repo_id} - check postsubmit.yml configuration"
+            exit 1
```
mering, line 40: With this a malicious actor could purposely block releases for several days. Consider requesting more results from the API or iterating over it.

drigz: I considered this but they could still block releases if we request more than one (it's hard for us to put a bound on how many malicious action runs they could trigger). Also the `event=schedule` limit should protect us from the DoS anyway.

drigz:
@mering PTAL - I also updated the description. I decided against `git branch --contains` because it seemed to duplicate the existing checks - and might be annoying if we start maintaining release branches.

.github/workflows/release.yml
mering: Oh indeed, with the `event=schedule` this should always return a valid result. So reaching that code path should never happen but it is a good additional layer of protection (in case someone changes the query in the future when a nightly release might no longer be sufficient).

mering:
> @mering PTAL - I also updated the description. I decided against `git branch --contains` because it seemed to duplicate the existing checks - and might be annoying if we start maintaining release branches.

I agree that the `branch=main` query filter together with `head_repository.id` check would be at least equivalent to the `git branch --contains` (probably even superior).

mering (approved):

PR #388, opened 2024-07-16
cert-manager: bump 1.14.2 -> 1.15.1

therjak:


ensonic (approved):

therjak:
currently this only shows the need to update the helm library used to helm3. there was a major refactoring between helm2 and helm3 so it will take some time.

PR #391, opened 2024-07-22
Fix typo and log noise.

ensonic:


scripts/robot-sim.sh
```diff
@@ -16,8 +16,8 @@
 
 # Manage simulated robots
 #
-# Simulated robots are does as a separate cluster, running the same components like a physical robot
-# in addition to the robot simulator.
+# Simulated robots are done as a separate cluster, running the same components
```
csieber, line 19: "done" sounds a bit strange here, maybe "implemented"?

csieber (approved):

scripts/robot-sim.sh
ensonic: done.

PR #396, opened 2024-07-29
Allow to configure CRC to use regional clusters

ensonic:
So far we always created "zonal" clusters. This set of changes introduced a new config.sh var `GKE_CLUSTER_TYPE` that defaults to "zonal". This var will be feed into terraforms input vars and allows to select create "regional" clusters instead.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -18,6 +18,7 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "${DIR}/scripts/common.sh"
+source "${DIR}/scripts/config.sh"
 source "${DIR}/scripts/include-config.sh"
 
 set -o pipefail -o errexit
@@ -57,6 +58,19 @@ function include_config_and_defaults {
   CLOUD_ROBOTICS_CTX=${CLOUD_ROBOTICS_CTX:-"gke_${GCP_PROJECT_ID}_${GCP_ZONE}_${PROJECT_NAME}"}
 
   SYNK="${SYNK_COMMAND} --context ${CLOUD_ROBOTICS_CTX}"
+}
+
+function update_config_var {
+  cloud_bucket="gs://${1}-cloud-robotics-config"
+  name="${2}"
+  value="${3}"
+
+  config_file="$(mktemp)"
+  gsutil cp "${cloud_bucket}/config.sh" "${config_file}" 2>/dev/null || return
+
+  save_variable "${config_file}" "${name}" "${value}"
+
+  gsutil mv "${config_file}" "${cloud_bucket}/config.sh"
 }
 
 function kc {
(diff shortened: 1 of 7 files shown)


deploy.sh
```diff
@@ -227,7 +227,23 @@ function terraform_apply {
...
+  fi
+  if [[ "${OLD_CLOUD_ROBOTICS_CTX}" != "${CLOUD_ROBOTICS_CTX}" ]]; then
+    echo "updating CLOUD_ROBOTICS_CTX from ${OLD_CLOUD_ROBOTICS_CTX} to ${CLOUD_ROBOTICS_CTX}"
+    # update CLOUD_ROBOTICS_CTX in config.sh
+    # TODO(ensonc): need to store the changes:
+    # a) extend set-config.sh to have --edit-var="key=value" ?
+    #   - would also need to add --verbose to supress the print of the new config ...
+    # b) duplicate code here? 
+    # bash -c "${DIR}/scripts/set-config.sh; --edit-var=CLOUD_ROBOTICS_CTX=${CLOUD_ROBOTICS_CTX}"
+  fi
```
ensonic, line 246: @drigz any (strong) opinion here?

drigz: No opinion, either seems fine.

ensonic:
Tested:
running `./deploy.sh update robco-ensonic` -> no change
added `GKE_CLUSTER_TYPE="regional"` to my config.sh and rerun prev cmd -> cluster is recreated as regional
change config.sh to `GKE_CLUSTER_TYPE="zonal"` and rerun prev cmd -> cluster is recreated as zonal

src/bootstrap/cloud/terraform/cluster.tf
```diff
@@ -103,7 +109,7 @@ resource "google_container_node_pool" "cloud_robotics_base_pool_ar" {
   for_each = var.additional_regions
   project  = data.google_project.project.project_id
   name     = format("%s-%s", "base-pool-ar", each.key)
-  location = each.value.zone
+  location = var.cluster_type == "zonal" ? each.value.zone : each.value.region
```
drigz, line 112: @Ongy Does it make sense to use regional clusters for the additional regions? Do we rely on high availability for the control plane or frontend there, or do we fall back to the US region? I think it's a difference of about $100/month/additional cluster, as we need a minimum of 3 nodes compared to our current 2.

drigz (approved):

src/bootstrap/cloud/terraform/cluster.tf
ensonic: @Ongy what do you think?

Ongy: I think for now it's fine if we keep them zonal, since we don't use the EU frontend much either way, and I'd like to get TURN onto something simpler than GKE either way.

OTOH, if we start using it more for some customer projects, I'd like to have the frontend in a reliable way.
and I think we might use it for at least one enterprise project soon? (I think this is public, so no names)

I'd say for now the simpler to implement/operate the transition is probably the better choice, and we see whether we upgrade  the additional ones later, resp. reduce/replace them if we don't need the features in a couple months.

ensonic: Removed.

PR #401, opened 2024-07-30
Avoid hanging after custom metrics are removed

### LINKED ISSUE #261: Bubbled up errors should cause the chart-assignment-controller to crash
I encountered an issue where a helm chart I deployed caused an error in the k8s `client-go` library to get bubbled up to the `chart-assignment-controller`.

The controller's logs would then display the error, but not crash or restart, instead freezing and stopping deployment of **all** new ChartAssignments.

In my case, the log was:
```
"couldn't get resource list for external.metrics.k8s.io/v1beta1: Got empty response for: external.metrics.k8s.io/v1beta1"
```

I'm not too sure where in the controller's code the error is caught, but I think it should result in a crash, so the breakage isn't hidden. Either that or it should ignore that error gracefully, but I'm not sure where the error is even bubbling up, unfortunately..

I have an inkling that it's related to the cached discovery client?
https://github.com/googlecloudrobotics/core/blob/main/src/go/pkg/synk/synk.go#L80-L94

drigz:
Previously, we had a hacky heuristic on when to trust partial discovery
information from the server, but this would still fail to apply
ChartAssignments when it got the error "stale GroupVersion discovery".
Instead, we can treat such errors as warnings, and instead look for
whether the resources in the charts are represented in the discovery
information.

Also, to reduce the frequency of the warnings, invalidate the discovery
cache before calling discovery - this means you just see the warning
once (due to a race in the apiserver I think) instead of the entire time
until the process is restarted.

Fixes #261. (well, we don't actually crash as that issue requests, but I like this way more)

### DIFF
--- a/src/go/pkg/synk/synk.go
+++ b/src/go/pkg/synk/synk.go
@@ -480,15 +480,21 @@ func (s *Synk) populateNamespaces(
 	resources ...*unstructured.Unstructured,
 ) error {
 	_, span := trace.StartSpan(ctx, "Discover server resources")
+	// Invalidate is cheap (no noticeable effect on the duration of
+	// ServerGroupsAndResources) and reduces the frequency of the "stale
+	// GroupVersion discovery" warning.
+	s.discovery.Invalidate()
 	_, list, err := s.discovery.ServerGroupsAndResources()
 	span.End()
 
-	// Don't treat "empty response" from external metrics servers as an error.
-	// Follows precedent from: https://github.com/kubernetes/client-go/commit/0bc91705fa50a531b8a1ee9dca3ef063ca71bb0c
 	if err != nil {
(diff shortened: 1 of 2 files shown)


src/go/pkg/synk/synk.go
```diff
@@ -511,7 +517,16 @@ func (s *Synk) populateNamespaces(
 		}
 	}
 	for _, r := range resources {
-		if r.GetNamespace() == "" && isNamespaced[r.GetAPIVersion()+"/"+r.GetKind()] {
+		if r.GetNamespace != "" {
```
therjak, line 520: r.GetNamespace()

drigz:
PTAL

therjak (approved):

PR #403, opened 2024-07-31
Delete old scripts.

ensonic:
I have not used them in ages and they probably do more harm that any good.


drigz (approved):

ensonic:
Had to also drop he script from two build targets. `git grep "show-resource-usage"` shows no reference though.

drigz (approved):

PR #405, opened 2024-08-01
Add a way to opt out of status checking

vivekbagade:
The label cloudrobotics.com/opt-out-error-checking=true when set on a pod will opt it out of the chartassignment controller's error checking routine.

### DIFF
--- a/src/go/pkg/controller/chartassignment/controller.go
+++ b/src/go/pkg/controller/chartassignment/controller.go
@@ -43,8 +43,8 @@ const (
 	// Allow the Service Account Controller some time to create the default
 	// SA in a new namespace.
 	defaultServiceAccountDeadline = time.Minute
-
-	fieldIndexNamespace = "spec.namespaceName"
+	fieldIndexNamespace           = "spec.namespaceName"
+	statusCheckingOptOutLabel     = "cloudrobotics.com/opt-out-error-checking"
 )
 
 // Add adds a controller and validation webhook for the ChartAssignment resource type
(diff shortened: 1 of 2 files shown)


ensonic:
Could you please give a little more context? Also please document the label, maybe here
https://github.com/googlecloudrobotics/core/blob/main/docs/concepts/app-management.md

src/go/pkg/controller/chartassignment/controller.go
```diff
@@ -394,8 +394,11 @@ func (r *Reconciler) setStatus(ctx context.Context, as *apps.ChartAssignment) er
 		// Determine readiness based on pods in the app namespace being ready.
 		// This is an incomplete heuristic but it should catch the vast majority of errors.
 		var pods core.PodList
+		// Omit pods that have opted out of status checking.
+		omitLabels := map[string]string{statusCheckingOptOutLabel: "!true"}
+
 		// Note, this return 0 is the namespace has not been created!
-		if err := r.kube.List(ctx, &pods, kclient.InNamespace(as.Spec.NamespaceName)); err != nil {
+		if err := r.kube.List(ctx, &pods, kclient.InNamespace(as.Spec.NamespaceName), kclient.MatchingLabels(omitLabels)); err != nil {
```
drigz, line 401: I think this will not match any pods.

The dev workflow for this repository is not well documented, so in hindsight I regret sending you down this rabbit hole, but the way I tested is:

```
bazel run src/app_charts/base/base-robot.push-all-containers.chart-assignment-controller.push -- --repository gcr.io/giza-workcells/chart-assignment-controller:$USER
# copy the SHA: gcr.io/giza-workcells/chart-assignment-controller:rodrigoq@sha256:02aad96674cfc04c960889988b7dc2d9b01e2127d482f5038a27f6bb68067e34 in my case
inctl vmpool lease --org intrinsic@giza-workcells --duration 30m --retry --context-alias vmkube
k9s --context vmkube
# :deploy, select chart-assignment-controller, `e` for edit, find `image:`, replace with fresh-built image
# then :chartassignment, select intrinsic-base, `y` for YAML, observe `message: 0/0 pods are running or succeeded`
```

I'm not sure how to fix this, maybe with labels.Parse("cloudrobotics.com/opt-out-error-checking notin (true)") or by moving the check to the loop instead of using a label selector?

vivekbagade: Thanks. Apparently, kubectl get pods -napp-intrinsic-base --context minikube --selector=intrinsic.ai/opt-out-error-checking!=true does not work in code.

Added an explicit filter, tested with the steps you put down. It works

vivekbagade:
> Could you please give a little more context? Also please document the label, maybe here https://github.com/googlecloudrobotics/core/blob/main/docs/concepts/app-management.md

Done

src/go/pkg/controller/chartassignment/controller.go
```diff
@@ -394,13 +394,22 @@ func (r *Reconciler) setStatus(ctx context.Context, as *apps.ChartAssignment) er
 		// Determine readiness based on pods in the app namespace being ready.
 		// This is an incomplete heuristic but it should catch the vast majority of errors.
 		var pods core.PodList
+
```
ensonic, line 397: remove

ensonic (approved):

drigz (approved):

PR #406, opened 2024-08-12
Update service enablement in terraform.

ensonic:
* replace gcr by gar
* drop pubsub
* add gcs

Use the compact form with a foreach.

### DIFF
--- a/src/bootstrap/cloud/terraform/address.tf
+++ b/src/bootstrap/cloud/terraform/address.tf
@@ -2,7 +2,7 @@ resource "google_compute_address" "cloud_robotics" {
   name       = "cloud-robotics"
   project    = data.google_project.project.project_id
   region     = var.region
-  depends_on = [google_project_service.compute]
+  depends_on = [google_project_service.project-services["compute"]]
 }
 
 resource "google_compute_address" "cloud_robotics_ar" {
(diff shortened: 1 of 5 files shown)


src/bootstrap/cloud/terraform/address.tf
```diff
@@ -2,13 +2,13 @@ resource "google_compute_address" "cloud_robotics" {
   name       = "cloud-robotics"
   project    = data.google_project.project.project_id
   region     = var.region
-  depends_on = [google_project_service.compute]
+  depends_on = [google_project_service.project-services["compute"]]
```
drigz, line 5: Should these names be fully-qualified? How does it map "compute" to "compute.googleapis.com" now that "compute" no longer appears literally in project.tf?

ensonic: You are right. Interestingly terraform did not complain, but the full key is clearly visible from the plan output:
```
google_project_service.project-services["endpoints.googleapis.com"]: Refreshing state... [id=robco-ensonic/endpoints.googleapis.com]
google_project_service.project-services["dns.googleapis.com"]: Refreshing state... [id=robco-ensonic/dns.googleapis.com]
google_project_service.project-services["compute.googleapis.com"]: Refreshing state... [id=robco-ensonic/compute.googleapis.com]
...
```

drigz (approved):

PR #409, opened 2024-08-26
Allow forward authentication with robot JWT

Ongy:
This allows users to configure proxies (e.g. nginx) to defer authentication to the token vendor not just via the OIDC token, but already with the JWT. Providing this method allows to authenticate "from the first request" without having to first allow traffict to the token vendor from the client. Which is useful to allow an authenticated proxy that's on the route between the client and the token-vendor.

TESTED=auto, not yet in practice

### DIFF
--- a/src/go/cmd/token-vendor/api/v1/v1.go
+++ b/src/go/cmd/token-vendor/api/v1/v1.go
@@ -255,6 +255,45 @@ func (h *HandlerContext) tokenOAuth2Handler(w http.ResponseWriter, r *http.Reque
 	w.Write(tokenBytes)
 }
 
+// Handle requests to verify if a given JWT is valid
+//
+// This handler provides a backend for common proxy servers forward authentication.
+// I.e. it allows robots to sign a JWT and present it to nginx for authentication.
+//
+// Method: GET
+// Headers:
+// - Authorization: Bearer ...
+func (h *HandlerContext) verifyJWTHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		api.ErrResponse(w, http.StatusMethodNotAllowed,
+			fmt.Sprintf("method %s not allowed, only %s", r.Method, http.MethodGet))
+		return
+	}
+
+	authHeader, ok := r.Header["Authorization"]
+	if !ok {
+		api.ErrResponse(w, http.StatusBadRequest,
+			"request did not provide Authorization header")
+		return
+	}
+
+	if len(authHeader) != 1 {
+		api.ErrResponse(w, http.StatusBadRequest,
+			fmt.Sprintf("%q auth headers provided. Only 1 allowed", len(authHeader)))
+		return
+	}
+
+	// Be slightly permissive here. Allow both forms
+	// Authorization: Bearer ...
+	// Authorization: ...
+	jwtString := strings.TrimPrefix(authHeader[0], "Bearer ")
+
+	if _, err := h.tv.ValidateJWT(r.Context(), jwtString); err != nil {
+		api.ErrResponse(w, http.StatusForbidden, "JWT not valid")
+		return
+	}
+}
+
 // Handle requests to verify if a given token has cloud access.
 //
 // The token is verified by testing if the token has `iam.serviceAccounts.actAs`
@@ -389,6 +428,7 @@ func Register(tv *app.TokenVendor, prefix string) error {
 	http.HandleFunc(path.Join(prefix, "public-key.publish"), h.publicKeyPublishHandler)
 	http.HandleFunc(path.Join(prefix, "token.oauth2"), h.tokenOAuth2Handler)
 	http.HandleFunc(path.Join(prefix, "token.verify"), h.verifyTokenHandler)
+	http.HandleFunc(path.Join(prefix, "jwt.verify"), h.verifyJWTHandler)
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


src/go/cmd/token-vendor/app/tokenvendor.go
```diff
@@ -95,33 +95,42 @@ func (tv *TokenVendor) GetOAuth2Token(ctx context.Context, jwtk string) (*tokens
...
 	if err := acceptedAudience(p.Aud, tv.accAud); err != nil {
-		return nil, errors.Wrapf(err, "validation of JWT audience failed (iss: %q)", p.Iss)
+		return "", errors.Wrapf(err, "validation of JWT audience failed (iss: %q)", p.Iss)
 	}
 	if !IsValidDeviceID(p.Iss) {
-		return nil, fmt.Errorf("missing or invalid device identifier (`iss`: %q)", p.Iss)
+		return "", fmt.Errorf("missing or invalid device identifier (`iss`: %q)", p.Iss)
 	}
 	deviceID := p.Iss
 	pubKey, err := tv.repo.LookupKey(ctx, deviceID)
```
drigz, line 115: Not blocking this CL, but this is another scaling concern: This lookup isn't cached so every websocket connection will eat into our GKE QPS.

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -255,6 +255,45 @@ func (h *HandlerContext) tokenOAuth2Handler(w http.ResponseWriter, r *http.Reque
...
+		return
+	}
+
+	// Be slightly permissive here. Allow both forms
+	// Authorization: Bearer ...
+	// Authorization: ...
+	jwtString := strings.TrimPrefix(authHeader[0], "Bearer ")
+
+	if _, err := h.tv.ValidateJWT(r.Context(), jwtString); err != nil {
+		api.ErrResponse(w, http.StatusForbidden, "JWT not valid")
```
drigz, line 292: slog.Error()?

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -389,6 +428,7 @@ func Register(tv *app.TokenVendor, prefix string) error {
 	http.HandleFunc(path.Join(prefix, "public-key.publish"), h.publicKeyPublishHandler)
 	http.HandleFunc(path.Join(prefix, "token.oauth2"), h.tokenOAuth2Handler)
 	http.HandleFunc(path.Join(prefix, "token.verify"), h.verifyTokenHandler)
+	http.HandleFunc(path.Join(prefix, "jwt.verify"), h.verifyJWTHandler)
```
drigz, line 431: Please also document here: https://github.com/googlecloudrobotics/core/blob/5929ab84000fb3c5d3e159686e2dafa4d6f02836/src/go/cmd/token-vendor/README.md?plain=1#L79

drigz, line 431: Do you need to extend https://github.com/googlecloudrobotics/core/blob/5929ab84000fb3c5d3e159686e2dafa4d6f02836/src/app_charts/token-vendor/cloud/ingress.yaml#L60?

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -255,6 +255,45 @@ func (h *HandlerContext) tokenOAuth2Handler(w http.ResponseWriter, r *http.Reque
 	w.Write(tokenBytes)
 }
 
+// Handle requests to verify if a given JWT is valid
+//
+// This handler provides a backend for common proxy servers forward authentication.
```
drigz, line 260: I feel this will be confusing to future readers who don't have the context you have now, especially the noun phrase "common proxy servers forward authentication". Does that refer to the nginx auth-url webhook? Or to another proxy?

Should we just copy-paste the docs from above and avoid documenting how this is used here? That is to say,  #copying:

```
// The robot identifies itself using its private key to sign a JWT. The JWT is
// verified by the token vendor using the public key from the key repository and
// the device identifier from the `iss` key from the JWT's payload section.
```

And explain that this can be used as an authentication check but that it doesn't apply authz beyond comparing to the list of keys in the key repository?

src/go/cmd/token-vendor/api/v1/v1.go
Ongy: This happens on every validation error, including expired. So at most I'd say Warn.

Might be interesting to log some internal ones at a higher level, I'll take a look.

drigz: I'm not so fussed about the error, and this is indeed pretty noisy already from other handlers. An alternative would be able to generate metrics for the different error codes. My main concern is about having a way to debug unexpected failures.

src/go/cmd/token-vendor/api/v1/v1.go
Ongy: Yes.

src/go/cmd/token-vendor/api/v1/v1.go
Ongy: I replaced it by a copy+paste from the readme, since that seems like an reasonable explanation to me.

drigz:
Thanks! Realized Christian should probably have a chance to review this too.

drigz (approved):

PR #410, opened 2024-08-26
Add JWT signing utility to cloudrobotics library

Ongy:


csieber (changes requested):
Discussion: https://chat.google.com/room/AAAAFoy0l8M/7rESu8KpUHo/7rESu8KpUHo?cls=10

src/go/pkg/robotauth/robotauth.go
```diff
@@ -150,3 +169,40 @@ func (r *RobotAuth) CreateRobotTokenSource(ctx context.Context) oauth2.TokenSour
...
+		}
+	}
+	parsed, ok := parsedKey.(*rsa.PrivateKey)
+	if !ok {
+		return "", fmt.Errorf("private key is invalid")
+	}
+
+	claimSet := &jws.ClaimSet{
+		Iss:   r.PublicKeyRegistryId,
+		Scope: r.getTokenEndpoint(),
```
drigz, line 195: This is /apis/core.token-vendor/v1/token.oauth2, should it be jwt.verify instead? I don't know what the exact intent of scope/aud is, I'm just pattern matching on the strings.

drigz (commented):
IIUC CreateJWT() is required anyway, but we could either add the new path to the metadata-server in this or a followup PR.

src/go/cmd/metadata-server/metadata.go
```diff
@@ -105,6 +106,62 @@ func (ch ConstHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	i := &IdentityHandler{
+		AllowedSources: allowedSources,
+		robotAuth:      robotAuth,
+	}
+	return i, nil
+}
+
+func (h *IdentityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ipPort := strings.Split(r.RemoteAddr, ":")
```
csieber, line 137: Can we factor the RemoteAddr check out to a new function?

csieber:
Can you split the change into two? One for the metadata-server and one where you use it?

Ongy:
I'm not quite sure  which split you mean.

Currently the metadata-server part is the user of the new API. Using this is out of scope of this repo (the other PR with the new validation endpoint is the closest to user here).
I can split the new API (~the first commit) from the metadata-server part.

csieber:
> I'm not quite sure which split you mean.
> 
> Currently the metadata-server part is the user of the new API. Using this is out of scope of this repo (the other PR with the new validation endpoint is the closest to user here). I can split the new API (~the first commit) from the metadata-server part.

True. I misunderstood the change. Splitting it as you recommended would be good.

src/go/pkg/robotauth/robotauth.go
Ongy: While I think this would be technically better, I'm shying away from having a second audience on the token-vendor side, since it's currently provided as cmdline arg.

There's no access ability difference on the metadata-server, so I don't think we have a real gain from splitting it.
Though I can still do it for cleanliness.

drigz: Thanks, please leave a comment then to explain that it's an intentional simplification as it doesn't affect the use.

drigz (approved):

csieber (approved):

PR #414, opened 2024-09-02
Enable DNS01 challenges for cert-manager

drigz:
These only work when we provide a custom domain, but they work
regardless of what load balancer you use.


Ongy (approved):

ensonic (approved):
thanks for coming up with more narrow privileges.

PR #416, opened 2024-09-05
Allow to start metadata-server without internet connection

Ongy:
The previous approach got the metadata server stuck and prevented the setup of the iptables rule With this change, the metadata server can setup and only reject requests it cannot service due to the missing information

Using atomic operations since the project 0 should never show up and it is
* easier than proper locking
* does not lead to congestion on the read path we cannot wait for the resolution, but assume downstream clients retry

### DIFF
--- a/src/go/cmd/metadata-server/metadata.go
+++ b/src/go/cmd/metadata-server/metadata.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff"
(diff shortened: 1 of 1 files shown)


drigz:
To get it to the metadata-server we'd need to provide it when setting up the IPC cluster: https://github.com/intrinsic-ai/xfa-tools/blob/ff59de0614a714b4d988de0b49d8a2a5e4c0c8af/internal/cloudrobotics/cloudrobotics.go#L377

So it would either need to be plumbed/backfilled into config.yaml or stored somewhere else by the IPC when it accepts its configuration. For offline IPCs we'd need automated code to backfill. So I'd probably stick with this approach here as we don't have to introduce any new variation from one IPC to another.

drigz (approved):

ensonic:
> To get it to the metadata-server we'd need to provide it when setting up the IPC cluster: https://github.com/intrinsic-ai/xfa-tools/blob/ff59de0614a714b4d988de0b49d8a2a5e4c0c8af/internal/cloudrobotics/cloudrobotics.go#L377
> 
> So it would either need to be plumbed/backfilled into config.yaml or stored somewhere else by the IPC when it accepts its configuration. For offline IPCs we'd need automated code to backfill. So I'd probably stick with this approach here as we don't have to introduce any new variation from one IPC to another.

Good point. In the long run I still think it would be more robust if we could statically set this up. But I agree it is hard to prioritize.

PR #421, opened 2024-09-16
Remove @cloud_robotics prefix.

ensonic:
Address the error from the migration tool:
```
Please remove the usages of referring your own repo via `@cloud_robotics//`, targets should be referenced directly with `//`.
```

### DIFF
--- a/bazel/app_chart.bzl
+++ b/bazel/app_chart.bzl
@@ -89,7 +89,7 @@ _app_chart_backend = rule(
             doc = "Images referenced by the chart.",
         ),
         "_chart_yaml_template": attr.label(
-            default = Label("@cloud_robotics//bazel/build_rules/app_chart:Chart.yaml.template"),
+            default = Label("//bazel/build_rules/app_chart:Chart.yaml.template"),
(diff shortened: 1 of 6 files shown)


bazel/app_chart.bzl
```diff
@@ -134,9 +134,9 @@ def app_chart(
 
     if not values:
         if chart == "cloud":
-            values = "@cloud_robotics//bazel/build_rules/app_chart:values-cloud.yaml"
+            values = "//bazel/build_rules/app_chart:values-cloud.yaml"
         else:
-            values = "@cloud_robotics//bazel/build_rules/app_chart:values-robot.yaml"
+            values = "//bazel/build_rules/app_chart:values-robot.yaml"
```
mering, line 139: Use `Label()` in `.bzl` files for targets in this repo to resolve correctly when macro is called from a foreign repo.

mering (approved):

bazel/app_chart.bzl
ensonic: Done.

PR #426, opened 2024-09-17
Start to apply the layer 1 customization.

ensonic:
* make the -robot bucket optional
* make the robot-service account optional

I wanted to also avoid the token-vendor rollout in base-cloud, but that is related to human-acl and oauth2-proxy usage.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -157,7 +157,7 @@ robot_image_reference = "${SOURCE_CONTAINER_REGISTRY}/setup-robot@${ROBOT_IMAGE_
 crc_version = "${CRC_VERSION}"
 certificate_provider = "${CLOUD_ROBOTICS_CERTIFICATE_PROVIDER}"
 cluster_type = "${GKE_CLUSTER_TYPE}"
-onprem_federation = "${ONPREM_FEDERATION}"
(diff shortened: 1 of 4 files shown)


ensonic:
I tested this on robco-ensonic:

1. I redeployed not having ONPREM_FEDERATION set in my config.sh and verified that nothing gets deleted.
2. I've added the ONPREM_FEDERATION=false to my config.sh and redepled. The robco-ensonic-robot bucket and the robbot-service SA were deleted.
3. I've change the ONPREM_FEDERATION to true and redeployed. The resources were recreated.

ensonic:
I had to rewrite registry.tf. Terraform is not good dealing with variable size data-structures ...

ensonic:
re-tested this on robco-ensonic:

1. I redeployed not having ONPREM_FEDERATION set in my config.sh and verified that nothing gets deleted.
2. I've added the ONPREM_FEDERATION=false to my config.sh and redeployed. The robco-ensonic-robot bucket and the robot-service SA were deleted.
3. I've change the ONPREM_FEDERATION to true and redeployed. The resources were recreated.

Major difference, with the restructured code in `registry.tf` I can omit the SA.

Ongy (approved):

PR #428, opened 2024-09-18
More debugging in deploy.sh

ensonic:
Breakage is before reelase runs.


therjak (approved):

ensonic:
Thanks, I'll vert this once everything works again.

PR #435, opened 2024-09-18
Remove debug log output from ci

ensonic:


deploy.sh
```diff
@@ -108,13 +108,6 @@ function prepare_source_install {
     echo "bazel-bin: --------------------"
   fi
```
mering, line 109: You probably want to remove this debug output as well.

mering (approved):

deploy.sh
ensonic: Indeed. thanks!

PR #443, opened 2024-09-20
Restore comments which have been lost in the Bzlmod migration

mering:
We can now specify platform directly, so we use the manifest and platform now directly and the outdated comment for debian-iptables has been removed.

The `iptables_base` image doesn't have a latest tag: https://pantheon.corp.google.com/gcr/images/google-containers/GLOBAL/debian-iptables


ensonic (approved):
Thanks!

Issue #447, opened 2024-09-23
[Bazel CI] app_management_test is failing with Bazel@HEAD



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4142#01921d0e-ec6d-44af-a755-5c6640f93d9c

 Platform: Ubuntu

 Logs: 
```
FAIL: [0m//src/app_charts/base:app_management_test (Exit 1) (see /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e68609a51962c4af9517a3c40321ba14/execroot/_main/bazel-out/k8-fastbuild/testlogs/src/app_charts/base/app_management_test/test_attempts/attempt_1.log)
FAIL: [0m//src/app_charts/base:app_management_test (Exit 1) (see /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e68609a51962c4af9517a3c40321ba14/execroot/_main/bazel-out/k8-fastbuild/testlogs/src/app_charts/base/app_management_test/test_attempts/attempt_2.log)
FAIL: [0m//src/app_charts/base:app_management_test (Exit 1) (see /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/e68609a51962c4af9517a3c40321ba14/execroot/_main/bazel-out/k8-fastbuild/testlogs/src/app_charts/base/app_management_test/test.log)
FAILED: [0m//src/app_charts/base:app_management_test (Summary)
Error: The command exited with status 1[0m
error: The plugin docker command hook exited with status 1
``` 
 Culprit: 

 Steps:
  ``` 
 git clone https://github.com/googlecloudrobotics/core
git reset c66169cd7e5d27849c2e777dba7e4b6949e87768  --hard
export USE_BAZEL_VERSION=b084956e23a6e50e35fe8886d0f3f7e05f718155
bazel test //... 
 ``` 
 CC Greenteam @wyverald

ensonic:
Its green now:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/4150#01922c85-003c-40ee-aa6f-8487bd300c66

PR #448, opened 2024-09-23
Fix app_management_test at HEAD

drigz:
See #447. Tested with:

```
bazel test //src/app_charts/base:app_management_test
USE_BAZEL_VERSION=b084956e23a6e50e35fe8886d0f3f7e05f718155 bazel test //src/app_charts/base:app_management_test
```

### DIFF
--- a/src/app_charts/base/app_management_test.sh
+++ b/src/app_charts/base/app_management_test.sh
@@ -14,11 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-if [[ -x "${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm" ]] ; then
-  HELM="${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm template"
-elif [[ -x "${TEST_SRCDIR}/+non_module_deps+kubernetes_helm/helm" ]] ; then
+HELM="${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm"
+if [[ ! -x "${HELM}" ]] ; then
   HELM="${TEST_SRCDIR}/+non_module_deps+kubernetes_helm/helm template"
-else
+fi
+if [[ ! -x "${HELM}" ]] ; then
   # If we hit this again, consider using the runfiles library:
   # https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash#L55-L86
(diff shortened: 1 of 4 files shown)


src/app_charts/base/app_management_test.sh
```diff
@@ -14,7 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-HELM="${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm template"
+if [[ -x "${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm" ]] ; then
+  HELM="${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm template"
+elif [[ -x "${TEST_SRCDIR}/+non_module_deps+kubernetes_helm/helm" ]] ; then
+  HELM="${TEST_SRCDIR}/+non_module_deps+kubernetes_helm/helm template"
+else
```
ensonic, line 21: can we do the same in deploy.sh? Any idea why my find approach failed? Btw. my find approach worked locally, but failed when I manually triggered the integration tests action workflow.

ensonic: Maybe less copy'n'paste if we write:
```
HELM="${TEST_SRCDIR}/_main/external/_main~non_module_deps~kubernetes_helm/helm"
if [[ ! -x "${HELM}" ]]; then
  HELM="${TEST_SRCDIR}/+non_module_deps+kubernetes_helm/helm template"
fi
if [[ ! -x "${HELM}" ]]; then
  echo >&2 "Failed to locate helm in ${TEST_SRCDIR}."
  exit 1
fi
```

drigz: Done. I copied it to deploy.sh, but I was too lazy to reimplement the logic in Go, so I just left a comment that hopefully someone will find when they update the Bazel pin.

ensonic (approved):

ensonic:
Sadly it fails on our CI, same as my change failed:
https://github.com/googlecloudrobotics/core/actions/runs/11034197473/job/30647145593

drigz:
Ah, thanks for the link. I think it's because the build hasn't happened at that point of deploy.sh. We should probably just keep it hardcoded there and update it when the pin changes.

ensonic:
Ahh, that's a bit silly. +1 for then just adding another TODO there. Would be much better to just set this further own in the script.

PR #449, opened 2024-09-25
Enable --incompatible_use_plus_in_repo_names

drigz:
This avoids the complexity of handling two different paths.
See #447.


ensonic (approved):
Ohh, that's much better. Thanks!

PR #450, opened 2024-10-01
Switch to artifact registry.

ensonic:
This was reenabling the contaienrregistry api in tests.


drigz (approved):
Ouch, nasty bit of complexity, thanks for the fix.

PR #454, opened 2024-10-11
Add updated lockfile.

ensonic:
We've udpated bazel and some entries changed.


drigz (approved):

drigz:
Thanks, this has been annoying but I've been neglecting to fix it.

ensonic:
> Thanks, this has been annoying but I've been neglecting to fix it.

same here ... - had to read a bit on whether that file is supposed to be submitted. Shoudl have added the reference: https://bazel.build/external/lockfile#best-practices

PR #460, opened 2024-11-22
API proposal for the custom sa associations.

ensonic:
See b/367635510

### DIFF
--- a/src/go/cmd/token-vendor/README.md
+++ b/src/go/cmd/token-vendor/README.md
@@ -54,6 +54,21 @@ refuse eg. robots to register other robots.
 * Body: application/x-pem-file
 * Response: only http status code
 
+### /public-key.configure: Customize robot registration
+
+Status: Proposed
+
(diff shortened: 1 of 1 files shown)


src/go/cmd/token-vendor/README.md
```diff
@@ -54,6 +54,21 @@ refuse eg. robots to register other robots.
 * Body: application/x-pem-file
 * Response: only http status code
 
+### /public-key.configure: Customize robot registration
```
csieber, line 57: Doing that only in retro-perspective sounds like there is a race-condition where for a short-time the robot-service account is used. Can we also add this at registering?

ensonic: It will happen when you bind the IPC

csieber: I think we should also try to add it to the publish call.

csieber (approved):

PR #461, opened 2024-11-22
Allow using a custom service account on an ipc-registration.

ensonic:
One can already configure a custom service account by using the annotation on the configmap. The api will be added in teh next changes.

### DIFF
--- a/src/go/cmd/token-vendor/app/tokenvendor.go
+++ b/src/go/cmd/token-vendor/app/tokenvendor.go
@@ -36,7 +36,7 @@ type PubKeyRepository interface {
 	// LookupKey retrieves the public key of a device from the repository.
 	// An empty string return indicates that no key exists for the given identifier or
 	// that the device is blocked.
-	LookupKey(ctx context.Context, deviceID string) (string, error)
+	LookupKey(ctx context.Context, deviceID string) (string, string, error)
 	PublishKey(ctx context.Context, deviceID, publicKey string) error
 }
 
@@ -61,7 +61,8 @@ func (tv *TokenVendor) PublishPublicKey(ctx context.Context, deviceID, publicKey
 
 func (tv *TokenVendor) ReadPublicKey(ctx context.Context, deviceID string) (string, error) {
 	slog.Debug("Returning public key", slog.String("DeviceID", deviceID))
-	return tv.repo.LookupKey(ctx, deviceID)
+	key, _, err := tv.repo.LookupKey(ctx, deviceID)
+	return key, err
 }
 
 var (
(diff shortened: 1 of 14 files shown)


ensonic:
Mostly FYI (and to save my changes).

src/go/cmd/token-vendor/app/tokenvendor.go
```diff
@@ -36,7 +36,7 @@ type PubKeyRepository interface {
 	// LookupKey retrieves the public key of a device from the repository.
 	// An empty string return indicates that no key exists for the given identifier or
 	// that the device is blocked.
```
csieber, line 38: Can we document the two return values here?

src/go/cmd/token-vendor/repository/k8s/k8s.go
```diff
@@ -63,21 +65,22 @@ func (k *K8sRepository) ListAllDeviceIDs(ctx context.Context) ([]string, error)
 // The public key is stored under a specific key in the configmap. If the configmap
 // does not exist, we return an empty string. For any other error or a malformed
 // configmap we return an error.
-func (k *K8sRepository) LookupKey(ctx context.Context, deviceID string) (string, error) {
+func (k *K8sRepository) LookupKey(ctx context.Context, deviceID string) (string, string, error) {
```
csieber, line 68: I think we should rename this to LookupConfig and return a config struct here. However that can be a follow-up.

src/go/cmd/token-vendor/tokensource/gcp.go
```diff
@@ -44,15 +44,19 @@ func NewGCPTokenSource(ctx context.Context, client *http.Client, project, sa str
 // Token returns an access token for the configured service account and scopes.
 //
 // API: https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken
-func (g *GCPTokenSource) Token(ctx context.Context) (*TokenResponse, error) {
+func (g *GCPTokenSource) Token(ctx context.Context, saName string) (*TokenResponse, error) {
 	req := iam.GenerateAccessTokenRequest{Scope: g.scopes}
+	resource := g.resource
```
csieber, line 49: can we rename resource to defaultResource or similar?

csieber: And match the name with the parameter name

csieber (approved):

src/go/cmd/token-vendor/app/tokenvendor.go
ensonic: I only uploaded the PR to save it. What if I introduce a struct Key that has all the values?

src/go/cmd/token-vendor/tokensource/gcp.go
ensonic: Done.

src/go/cmd/token-vendor/repository/k8s/k8s.go
ensonic: Done

src/go/cmd/token-vendor/app/tokenvendor.go
ensonic: Done

ensonic:
PTAL. Maybe as a followup I pass the defaultSA to the tokenvendor app and not the token source. Then the token source is just receiving the sa-name and getting a token for it. That also fixes the logging issue. WDYT?

src/go/cmd/token-vendor/repository/repository.go
```diff
@@ -0,0 +1,35 @@
...
+package repository
+
+import (
+	"context"
+)
+
+// Key holds data + metadata of a public key entry
+type Key struct {
+	PublicKey string
+	SAName    string
```
csieber, line 25: maybe document this as optional

csieber (approved):

src/go/cmd/token-vendor/app/tokenvendor.go
```diff
@@ -95,48 +96,48 @@ func (tv *TokenVendor) GetOAuth2Token(ctx context.Context, jwtk string) (*tokens
...
 	if err != nil {
 		return nil, err
 	}
-	cloudToken, err := tv.ts.Token(ctx)
+	cloudToken, err := tv.ts.Token(ctx, sa)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to retrieve a cloud token for device %q", deviceID)
 	}
-	slog.Info("Handing out cloud token", slog.String("DeviceID", deviceID))
+	slog.Info("Handing out cloud token", slog.String("DeviceID", deviceID), slog.String("ServiceAccount", sa))
```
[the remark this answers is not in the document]
csieber: The other option could be a function like `func (context.Context, string) (*TokenSource, error)`

csieber:
> PTAL. Maybe as a followup I pass the defaultSA to the tokenvendor app and not to the token source. Then the token source is just receiving the sa-name and getting a token for it. That also fixes the logging issue. WDYT?

Maybe two arguments. `func(contect.Context, string) (*TokenSource, error)` and default service account.

PR #462, opened 2024-11-25
Make tokensource less stateful.

ensonic:
Now it always gets the service-account-name. Putting the amail together has been move up to main.

This way we can log the used service accoutn in one place.

### DIFF
--- a/src/go/cmd/token-vendor/api/v1/v1_test.go
+++ b/src/go/cmd/token-vendor/api/v1/v1_test.go
@@ -27,6 +27,7 @@ import (
 const (
 	testPubKey    = "testdata/rsa_cert.pem"
 	jwtBodyPrefix = "grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion="
+	saName        = "robot-service@testproject.iam.gserviceaccount.com"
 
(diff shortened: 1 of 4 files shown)


src/go/cmd/token-vendor/api/v1/v1_test.go
```diff
@@ -129,7 +129,7 @@ func runPublicKeyReadHandlerWithK8sCase(t *testing.T, test *publicKeyReadHandler
 		t.Fatal(err)
 	}
 	// Setup app & API handler
-	tv, err := app.NewTokenVendor(context.TODO(), kcl, nil, nil, "aud")
+	tv, err := app.NewTokenVendor(context.TODO(), kcl, nil, nil, "aud", "robbot@project.iam.gserviceaccount.com")
```
csieber, line 132: robbot -> robot

csieber: and maybe make it a constant

csieber (approved):

src/go/cmd/token-vendor/api/v1/v1_test.go
ensonic: done for both

PR #464, opened 2024-11-26
Extend run-install to allow only applying terraform config.

ensonic:
When restoring a backup, it is useful to run the individual install steps.

See b/323148080

### DIFF
--- a/src/bootstrap/cloud/run-install.sh
+++ b/src/bootstrap/cloud/run-install.sh
@@ -30,12 +30,14 @@ else
   COMMAND="$3"
 fi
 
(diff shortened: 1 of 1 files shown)


ensonic:
I have not tested that, only carefully read over it again and again. My best idea so far ist to try it tomorrow on e.g. x-a-a. Any better idea?

drigz:
SGTM, I don't think this will break anyone. My only concern would be whether update_infra is sufficient to remove and recreate the cluster when changing zonal->regional but I can't think why it would fail.

drigz (approved):

ensonic:
> My only concern would be whether update_infra is sufficient to remove and recreate the cluster when changing zonal->regional but I can't think why it would fail.

It is going to run the terraform apply step. So it should work. I'll test this on x-a-a first without changint eh cofnig and tehn another day migrating it to regional.

PR #470, opened 2024-12-03
Upgrade terraform to v0.13.7.

ensonic:
This inbetween version is required as it introduces the provider config (see verisons.tf).


ensonic:
tested on robco-ensonic

faximan (approved):

PR #473, opened 2024-12-04
Import the config bucket and put installed version file there.

ensonic:
1) We reference the config bucket from a lot of the scripts and docs. Hence
   make sure we also create it here. Use terraform import cli, since we're
   still on a too old version for using the import statement.
2) create the crc version in the config bucket, as the setup-robot is
   conditional.


ensonic:
Tested on robco-ensonic

ensonic:
Sadly the cfg bucket is always in the us :/

faximan (approved):

PR #474, opened 2024-12-12
Add support for enabling secret manager plugin on cluster

Ongy:
This allows integration with the secret manager via kubernetes secrets. It allows to keep secrets secure while providing it to applications via file.

Current intended usecase is an auth_password_file for alertmanager.

TESTED=not yet

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -51,6 +51,7 @@ function include_config_and_defaults {
   CLOUD_ROBOTICS_DOMAIN=${CLOUD_ROBOTICS_DOMAIN:-"www.endpoints.${GCP_PROJECT_ID}.cloud.goog"}
   APP_MANAGEMENT=${APP_MANAGEMENT:-false}
   ONPREM_FEDERATION=${ONPREM_FEDERATION:-true}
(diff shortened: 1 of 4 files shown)


deploy.sh
```diff
@@ -429,6 +431,7 @@ function helm_region_shared {
     --set-string "oauth2_proxy.client_secret=${CLOUD_ROBOTICS_OAUTH2_CLIENT_SECRET}"
     --set-string "oauth2_proxy.cookie_secret=${CLOUD_ROBOTICS_COOKIE_SECRET}"
     --set "use_tv_verbose=${CRC_USE_TV_VERBOSE}"
+    --set "secret_manager_plugin=${GKE_SECRET_MANAGER_PLUGIN}"
```
drigz, line 434: This line is not needed/useful if you don't add it to values.yaml for the base cloud chart.

drigz (approved):

ensonic (approved):

deploy.sh
ensonic: +1. But telling charts whether it is enabled or not might be useful though.

PR #475, opened 2024-12-14
Use the member field for better readbility.

ensonic:
See https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account#attributes-reference


drigz (approved):
Nice cleanup, didn't know about this one. Thanks!

PR #482, opened 2025-01-29
Cleanup robotauth tokensource config.

ensonic:
Leave the scopes empty
This is optional and we already sent the data else where. I've deployed this to robco-ensonic and connected minikube to it. I also started hello-k8s and observed that the logs are captured in the cloud.

Add some more comments on the other fields as well.

### DIFF
--- a/src/go/pkg/robotauth/robotauth.go
+++ b/src/go/pkg/robotauth/robotauth.go
@@ -160,12 +160,14 @@ func (r *RobotAuth) getTokenEndpoint() string {
 // service account.
 func (r *RobotAuth) CreateRobotTokenSource(ctx context.Context) oauth2.TokenSource {
 	c := jwt.Config{
-		Email:      r.PublicKeyRegistryId, // Will be used as "issuer" of the outgoing JWT.
+		// Will be used as "issuer" of the outgoing JWT. Is not formatted as an email though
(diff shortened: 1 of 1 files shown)


src/go/pkg/robotauth/robotauth.go
```diff
@@ -160,12 +160,14 @@ func (r *RobotAuth) getTokenEndpoint() string {
...
-		Email:      r.PublicKeyRegistryId, // Will be used as "issuer" of the outgoing JWT.
+		// Will be used as "issuer" of the outgoing JWT. Is not formatted as an email though
+		Email:      r.PublicKeyRegistryId,
 		Expires:    time.Minute * 30,
 		PrivateKey: r.PrivateKey,
-		Scopes:     []string{r.getTokenEndpoint()},
-		Subject:    r.RobotName,
-		TokenURL:   r.getTokenEndpoint(),
+		Scopes:     []string{},
+		// TODO: shouldn't this be the service-account name we want to get the token for?
```
Ongy, line 168: iirc. this is used by the token vendor to find the correct private key.

Ongy (approved):

Ongy:
Yea, the fact that this isn't used is what caused me to write the interceptor in xfa-advertise instead of copying out this part.

PR #484, opened 2025-02-07
Add the log_level flag to relay server and set the default to WARN

ensonic:
This aligns the server with the code in the client and reduces log noise for a working system.

See b/393312849


ensonic:
Tested on robco-ensonic

therjak (approved):

PR #485, opened 2025-03-04
Protect PutRequestStream with a mutex

koonpeng:
`pr.requestStream` may be closed while it is writing because there is no mutex protecting it. This can happen when the request is reaped due to inactivity, if this happens then the server will panic.

### DIFF
--- a/src/go/cmd/http-relay-server/server/broker.go
+++ b/src/go/cmd/http-relay-server/server/broker.go
@@ -319,7 +319,7 @@ func (r *broker) ReapInactiveRequests(threshold time.Time) {
 	for id, pr := range r.resp {
 		if pr.lastActivity.Before(threshold) {
 			slog.Info("Timeout on inactive request", slog.String("ID", id))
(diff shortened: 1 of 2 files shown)


src/go/cmd/http-relay-server/server/broker.go
```diff
@@ -235,12 +237,21 @@ func (r *broker) GetRequestStream(id string) ([]byte, bool) {
 func (r *broker) PutRequestStream(id string, data []byte) bool {
 	r.m.Lock()
 	pr := r.resp[id]
+	pr.requestStreamMutex.Lock()
```
drigz, line 240: I think this needs to be after the pr == nil check. If so you'll need to bring the nil check inside the critical section and make sure r.m is unlocked on both paths similar to SendResponse().

koonpeng: 43fa715eb89c805f31db4ec610fea574e16ef732

PR #487, opened 2025-03-05
Add dependabot.yml

drigz:
actions/cache was not updated, leading to a presubmit failure. I hope
this makes dependabot also update GitHub Actions avoiding this in
future.

I've never configured Dependabot before and don't know how to test this
so apologies in advance if it creates many noisy PRs.


ensonic (approved):
Added Alex as well, as he might have done it in the past

PR #488, opened 2025-03-05
Run Gazelle to fix BUILD files

drigz:
This looks to be a no-op but makes it easier to keep BUILD dependencies
up to date with `bazel run :gazelle`.


ensonic (approved):
Thanks!

PR #493, opened 2025-03-06
Reset logName in fluentbit.

ensonic:
The entry is causing high cardinality errors in cloud logging and prevent the use of log based metrics.  The infomation is anyway already contained in dedicated fields. The stackdriver output plugin will then use a default value. GCP is applying a similar filtering.

See https://github.com/fluent/fluent-bit/issues/9897


ensonic:
I am confused. It does not consistently works:
https://cloudlogging.app.goo.gl/N6PLTsEPWwkcLAhU7
A few records are rewritten and a few are not. Do you have any idea?

therjak (approved):

PR #497, opened 2025-03-07
Add a log_level flag to all services.

ensonic:
Also mark existing "verbose" flag as deprecated.

See b/393312849


therjak (approved):

ensonic:
Meh, we should have called them `log-level` and not `log_level`. How much do we care, since we did not deploy, should I rename theses new ones and alias the old ones ?

therjak:
the change would not break anything as far as I can tell. So feel free to send a pr.

PR #498, opened 2025-03-10
Enable metadata-server on GKE

drigz:
This makes the GKE environment more similar to real systems, and in
particular will let the cr-syncer fetch robot JWTs from the
metadata-server.

Tested with:

```
./deploy.sh update my-test-project
./scripts/robot-sim.sh create my-test-project robot-sim
```

then checking that cr-syncer and metadata-server logs both looked OK.

### DIFF
--- a/src/go/cmd/metadata-server/main.go
+++ b/src/go/cmd/metadata-server/main.go
@@ -50,6 +50,7 @@ var (
 	minTokenExpiry = flag.Int("min_token_expiry", 300, "Minimum time a token needs to be valid for in seconds")
 	logPeerDetails = flag.Bool("log_peer_details", false, "When enabled details about the peer that requests ADC are logged on the expense of some extra latency")
 	logLevel       = flag.Int("log_level", int(slog.LevelInfo), "the log message level required to be logged")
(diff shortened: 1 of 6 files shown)


src/go/cmd/metadata-server/main.go
```diff
@@ -194,10 +195,12 @@ func main() {
...
-	if err := PatchCorefile(ctx, k8s); err != nil {
-		removeIPTablesRule()
-		slog.Error("PatchCorefile", slog.Any("Error", err))
-		os.Exit(1)
+	if !*runningOnGKE {
+		if err := PatchCorefile(ctx, k8s); err != nil {
+			removeIPTablesRule()
+			slog.Error("PatchCorefile", slog.Any("Error", err))
+			os.Exit(1)
+		}
```
ensonic, line 203: Do you also want to skip 'RevertCorefile(ctx, k8s)' with this condition?

drigz: Yes, thanks - a bit of copy-pasting but worth it to avoid generating errors on shutdown in GKE.

PR #504, opened 2025-03-13
Also apply cluster_type settings to AR clusters.

ensonic:
For deployments that choose to be reginal, the same reason applies to all their clusters.


ensonic:
The used config looks like this:
```
ADDITIONAL_REGIONS=('{"name": "europe-west1", "region": "europe-west1", "zone": "europe-west1-b"}')
```

Ongy (approved):

PR #519, opened 2025-04-11
Disable timeout handler

drigz:
It suppresses panic stacktraces:
https://github.com/golang/go/issues/27375

### DIFF
--- a/src/go/cmd/token-vendor/api/api.go
+++ b/src/go/cmd/token-vendor/api/api.go
@@ -50,8 +50,7 @@ func SetupAndServe(addr string) error {
 		Addr:         addr,
 		ReadTimeout:  httpTimeoutRead,
 		WriteTimeout: httpTimeoutWrite,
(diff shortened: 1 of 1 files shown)


drigz:
@csieber I don't know if the timeout handler has an important role here, should we submit this as-is or perhaps add a flag to turn it off?

csieber:
> @csieber I don't know if the timeout handler has an important role here, should we submit this as-is or perhaps add a flag to turn it off?

I guess without it the tv would crash the process on panic? I would leave it for now and add a flag to disable

drigz:
> I guess without it the tv would crash the process on panic? I would leave it for now and add a flag to disable

I don't believe so, when we disabled this we still saw the panic being gracefully caught by the http server: (full traceback is on the chat thread)

```
http: panic serving 10.88.3.28:56962: runtime error: invalid memory address or nil pointer dereference
```

PR #520, opened 2025-04-11
Refactor the repository to return NotFound for unknown keys

ensonic:
Previously we returned "" and later changed that to a nil struct. This causes panics, as we also don't return any error and without it dereference the returned value.

Define an error, return it and update the commetns accordingly.


csieber (approved):
thanks!

PR #521, opened 2025-04-11
Use informer to cache config maps in-memory.

drigz:
This should avoid outages due to the token-vendor hitting either the
client-side or server-side rate limit when clients ask for tokens too
frequently. I expect the next scaling limit will be the rate at which
nginx can handle the requests or the token-vendor can do the
cryptographic checks on the JWTs.

Tested by deploying to my project and seeing if it falls over - also
k8s.go has unit test coverage - I didn't have a good idea how to do load
testing.

Please review carefully: This code was mostly generated by Gemini but I
deleted an unnecessary mutex (the informer uses a thread-safe store) and
had to plumb the namespace into the configmap so we can update the
in-memory store when publishing keys.


src/go/cmd/token-vendor/repository/k8s/k8s.go
```diff
@@ -18,20 +18,32 @@ import (
...
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/googlecloudrobotics/core/src/go/cmd/token-vendor/repository"
+	"github.com/googlecloudrobotics/ilog"
 )
 
+// Re-list all ConfigMaps periodically in case we drop an event or otherwise
+// have inconsistent state in the cache. May be unnecessary but provide some
+// defence in depth.
+const resyncPeriod = 5 * time.Minute
```
ensonic, line 39: make it 10 min or even 1h. We've bumped that a while ago all over the place

src/go/cmd/token-vendor/repository/k8s/k8s.go
```diff
@@ -94,40 +131,43 @@ func (k *K8sRepository) LookupKey(ctx context.Context, deviceID string) (*reposi
...
 	slog.Debug("publishing key", slog.String("DeviceID", deviceID))
-	cm, err := createPubKeyDeviceConfig(deviceID, publicKey)
+	cm, err := createPubKeyDeviceConfig(deviceID, k.ns, publicKey)
 	if err != nil {
 		return errors.Wrapf(err, "failed to init device configmap %q/%q", k.ns, deviceID)
 	}
 	if _, err = k.kcl.CoreV1().ConfigMaps(k.ns).Create(ctx, cm, metav1.CreateOptions{}); err == nil { // no error
-
-		slog.Info("created new configmap", slog.String("Namespace", k.ns), slog.String("ConfigMap", deviceID))
+		k.cmInformer.GetStore().Add(cm)
```
ensonic, line 139: wouldn't the informcer figure that it has been added on its own?

src/go/cmd/token-vendor/repository/k8s/k8s.go
drigz: Done, thanks, I always wondered why that was so common. I set to an hour but considered removing (Gemini actually set it to zero originally but I was too scared).

drigz: Done, thanks, I always wondered why that was so common. I set to an hour but considered removing (Gemini actually set it to zero originally but I was too scared).

src/go/cmd/token-vendor/repository/k8s/k8s.go
drigz: Thanks, that's indeed inconsistent. I dug in and it only fails if we create a createPubKeyDeviceConfig returns a config map without a `metadata` property. But I added the warning and comment just in case to help readers.

Another possibility would be to use the cache as read-through and if we get "not found" then we read from k8s - it should work in practice but it'd break the assumptions of the PublishKey() test, and making that async seemed worse than just keeping the cache up-to-date here.

drigz:
> how important would it be for you do make this a new backend?

I think it's also safe to deploy the previous build if this change causes problems. The "CM not found" fix is not essential since the Golang HTTP server is smart enough to isolate the nil deref to the request and not kill the whole process. So I'll go ahead and merge this so we can push it to dev on Monday.

PR #523, opened 2025-04-14
Add a maintenance schedule for the cluster.

ensonic:
Configure the weekend for maintenance operations.

### DIFF
--- a/src/bootstrap/cloud/terraform/cluster.tf
+++ b/src/bootstrap/cloud/terraform/cluster.tf
@@ -16,7 +16,6 @@ resource "google_container_cluster" "cloud-robotics" {
 
   # Make the cluster VPC-native (default for v1.21+)
   networking_mode = "VPC_NATIVE"
-  ip_allocation_policy {}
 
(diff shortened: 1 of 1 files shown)


ensonic:
I can also try to be nice and make it dynamic based on the `prod` var.

drigz (approved):
Thanks! I guess there's no good way to test this in advance, but maybe you could leave  a reminder to check in ~4 weeks when the previous maintenance operations on the node pools were?

drigz:
>  I can also try to be nice and make it dynamic based on the prod var.

I think it would be better to move our dev/qa projects to a different release channel to detect issues earlier, otherwise we're too reliant on GKE's release schedule to avoid changes hitting prod before they hit dev.

ensonic:
> Thanks! I guess there's no good way to test this in advance, but maybe you could leave a reminder to check in ~4 weeks when the previous maintenance operations on the node pools were?

I'll add a TODO to my calendar to check https://console.cloud.google.com/kubernetes/upgrades/list for my personal dev-project weekly. The last update happened on a Thursday.

PR #524, opened 2025-04-15
Add a placeholder ingress for the cert.

ensonic:
This ensures that one ingress owns the tls-cert and avoids some repetition of the 'tls' block. The cert is determined through the 'host' field in the rules.

### DIFF
--- a/src/app_charts/base/cloud/cert-ingress.yaml
+++ b/src/app_charts/base/cloud/cert-ingress.yaml
@@ -1,0 +1,13 @@
+# Owns the 'tls' block to ensure a cert is generated and avoids repetition of 
+# the 'tls' block in other ingresses. Cert is associated via 'host' field.
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: certificate-ingress
+spec:
+  ingressClassName: nginx
+  tls:
+  - hosts:
+    - {{ .Values.domain }}
+  rules:
+  - {}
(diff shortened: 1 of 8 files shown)


ensonic:
I've tested this on my personal cloud project - you can check cloud-console for the ingress objects on the cluster).

I also ran:
```
openssl s_client -connect www.endpoints.robco-$USER.cloud.goog:443
curl -vvI https://www.endpoints.robco-$USER.cloud.goog/oauth2
```
Looking good so far. Anything else you'd test?

drigz:
> Looking good so far. Anything else you'd test?

Do you think it's worth deleting the GKE cluster to test this works for new projects too? I expect so as the nginx-ingress-controller seems mostly stateless, but that's the only thing I can think of that would go wrong here.

Please give @Ongy an FYI when you push this to intrinsic-portal-dev so we can test how that affects edge-dev.intrinsic.ai and coturn.portal-dev.intrinsic.ai (not sure that's the correct URL).

drigz (approved):

ensonic:
> Do you think it's worth deleting the GKE cluster to test this works for new projects too? I expect so as the nginx-ingress-controller seems mostly stateless, but that's the only thing I can think of that would go wrong here.

I've recreated my cluster, looks all good.

> Please give @Ongy an FYI when you push this to intrinsic-portal-dev so we can test how that affects edge-dev.intrinsic.ai and coturn.portal-dev.intrinsic.ai (not sure that's the correct URL).

I'll be OOO next week, so I would do this in last week of April or May

ensonic:
@Ongy FYI

PR #525, opened 2025-04-15
Run buildifier to add loads

comius:
Run `buildifier --lint=fix  -r -v .`. This adds loads for the rules that were removed from Bazel 8. By Bazel 9 automatic loads will be disabled and all loads will need to be explicitly present.

Remove @rules_proto dependency from MODULE.bazel. This has been replaced with @protobuf.

Add @rules_shell dependency to MODULE.bazel. This is needed for the used Shell rules.

See: https://github.com/bazelbuild/bazel/issues/25755


ensonic (approved):
Thanks for the cleanup!

PR #527, opened 2025-04-23
Revert "Add cleanup policy for artifact repositories."

csieber:
Reverts googlecloudrobotics/core#518

Removes images used in production.

chat: https://chat.google.com/room/AAAAjJIMTAM/pcWocoBr90E/pcWocoBr90E?cls=10


drigz:
I think we can probably roll this forward as soon as we fix the tagging but this is likely worth doing anyway. I'll look into disabling on intrinsic-assets-prod specifically in Pantheon. Might also need to tell Marco not to push CRC to prod next week.

drigz (approved):

csieber:
Integration tests fail: https://github.com/googlecloudrobotics/core/actions/runs/14633140909

Created google provider bug: https://github.com/hashicorp/terraform-provider-google/issues/22495

PR #530, opened 2025-05-05
Return 401 Unauthorized on missing header

drigz:
https://nginx.org/en/docs/http/ngx_http_auth_request_module.html - any response other than 401 or 403 is considered an error. However, this code path can easily be triggered by clients such as security scanners, leading to spurious "errors" in the nginx logs.


Ongy (approved):

ensonic (approved):

ensonic:
@drigz you'll need to also adjust the test.

ensonic (approved):

PR #532, opened 2025-05-06
Update .gitignore

ensonic:
Structure a bit and add new bazel files.


therjak (approved):

ensonic:
Given https://bazel.build/external/lockfile#best-practices I should probably drop this change (or not add MODULE.bazel.lock).

PR #537, opened 2025-05-27
Use typed client for ResourceSets

drigz:
This complicates the tests a little but avoids marshalling to/from JSON.


ensonic (approved):
Nice, thank you

PR #539, opened 2025-05-29
Allows TokenVendor to read SA from JWT.SUB

rastislav-vcrs:
Changes implementation of token vendor OAuth2 endpoit to read custom service account information from JWT.SUB instead from request body.

Callers cannot easily add extra information to JWT Token request body as common libraries do not allow to alter the resulting request from outside before it is sent to the server. This change ensures that callers can specify service account in a JWT Claim SUBJECT field which is then considered by TokenVendor when issuing a new token. This field was previously unused by TV and feels very natural to be provided without need for custom claims to be introduced. The device ID information is passed in ISS field of the JWT Request.

RobotAuth implementatiuon provides two code paths for generating JWT tokens for robot. One is used to exclusively communicate with TokenVendor, this one is not modified. The second path is used for obtaining GCP tokens. This path allows setting custom SA in subject claim.

### DIFF
--- a/src/go/pkg/robotauth/robotauth.go
+++ b/src/go/pkg/robotauth/robotauth.go
@@ -30,6 +30,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/googlecloudrobotics/core/src/go/pkg/kubeutils"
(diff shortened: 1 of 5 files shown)


drigz:
@rastislav-vcrs is it intentional that we've been added while it's in draft status? not clear if I should review / take a high-level glance / wait for status change.

src/go/pkg/robotauth/robotauth.go
```diff
@@ -158,24 +159,26 @@ func (r *RobotAuth) getTokenEndpoint() string {
...
 		PrivateKey: r.PrivateKey,
 		Scopes:     []string{},
-		// TODO: shouldn't this be the service-account name we want to get the token for?
-		Subject:  r.RobotName,
-		TokenURL: r.getTokenEndpoint(),
+		TokenURL:   r.getTokenEndpoint(),
+	}
+	if len(gcpSaChain) > 0 {
+		// TokenVendor expects single service-account email.
+		c.Subject = strings.Join(gcpSaChain, ",")
```
ensonic, line 175: This will be comma separated list of emails which contradicts the comment.

ensonic (commented):
Please also update the README
https://github.com/googlecloudrobotics/core/blob/main/src/go/cmd/token-vendor/README.md

Ongy (approved):

PR #545, opened 2025-06-18
Allow TokenVendor to read SA from jwt.SUB (reappky-fix)

rastislav-vcrs:
Adjusts the way token-vendor uses `jwt.SUB` as service account. Only
if value of `jwt.SUB` looks very like GCP Service Account it is used
as service account value during token provisioning. Otherwise,
empty value is returned to support legacy use-case where `jwt.SUB`
contained non-SA data.


ensonic (approved):
Thanks! Especially for the test.

PR #660, opened 2026-04-16
ensure responseChannel is drained

koonpeng:
There is a potential for deadlocks

https://github.com/googlecloudrobotics/core/blob/f173d97ba2477bc9080af9fc2e1d364cc9017173/src/go/cmd/http-relay-client/client/client.go#L708-L712

if we exit the loop preemptively, we fail to fully consume `responseChannel` and https://github.com/googlecloudrobotics/core/blob/f173d97ba2477bc9080af9fc2e1d364cc9017173/src/go/cmd/http-relay-client/client/client.go#L487

might deadlock.

#222 fixes the error handling which may have caused the deadlock. Before, it checks for incorrect `backoff.PermanentError` which never happens so the `responseChannel` is always fully consumed.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -412,7 +412,6 @@ func (c *Client) postResponse(remote *http.Client, br *pb.HttpResponse) error {
 	}
 
 	resp, err := remote.Post(responseUrl.String(), "application/vnd.google.protobuf;proto=cloudrobotics.http_relay.v1alpha1.HttpResponse", bytes.NewReader(body))
-
 	if err != nil {
 		return fmt.Errorf("couldn't post response to relay server: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


src/go/cmd/http-relay-client/client/client.go
```diff
@@ -708,6 +706,11 @@ func (c *Client) handleRequest(remote *http.Client, local *http.Client, pbreq *p
 		if err != nil {
 			slog.Error("Closing backend connection",
 				slog.String("ID", *resp.Id), ilog.Err(err))
+			// This is also closed in streamToBackend. In most cases this is safe but it depends on the transport.
```
ensonic: Can you add more detail for which transport this Close() call would be needed?

ensonic (approved):
Otherwise, thanks & +1

src/go/cmd/http-relay-client/client/client.go
koonpeng: The "default" http client doesn't panic. I didn't really tested if it would panic in our case.
In theory, we can also not close here. The server should eventually close the connection, then `streamBackend` will end up closing `hresp.Body`. Closing it here would just propagate the disconnect to the backend faster.

ensonic: Thanks maybe reword as:
// This is also closed in streamToBackend. Closing here too to ensure the disconnect propagates faster in case e.g. the default http client is used.

PR #662, opened 2026-04-22
Update tokenvendor logging.

ensonic:
Instead of producing 2 log messages in most cases log only from ErrResponse helper. Use a group to log the http error (as suggested by https://docs.cloud.google.com/storage/docs/json_api/v1/status-codes).

### DIFF
--- a/src/go/cmd/token-vendor/api/v1/v1.go
+++ b/src/go/cmd/token-vendor/api/v1/v1.go
@@ -75,38 +75,37 @@ func getQueryParam(u *url.URL, param string) (string, error) {
 func (h *HandlerContext) publicKeyConfigureHandler(w http.ResponseWriter, r *http.Request) {
 	// validate request and parameters
 	if r.Method != http.MethodPost {
-		api.ErrResponse(w, http.StatusBadRequest,
+		api.ErrResponse(r.Context(), w, http.StatusBadRequest,
 			fmt.Sprintf("method %s not allowed, only %s", r.Method, http.MethodPost))
 		return
 	}
 	deviceID, err := getQueryParam(r.URL, paramDeviceID)
 	if err != nil {
-		api.ErrResponse(w, http.StatusBadRequest, err.Error())
+		api.ErrResponse(r.Context(), w, http.StatusBadRequest, err.Error())
 		return
 	}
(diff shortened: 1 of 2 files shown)


ensonic:
Example, see last log-entry in https://cloudlogging.app.goo.gl/jQ3Dmqs2w1R7L9Ai8

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -75,38 +75,37 @@ func getQueryParam(u *url.URL, param string) (string, error) {
...
 	}
 	if !app.IsValidDeviceID(deviceID) {
-		api.ErrResponse(w, http.StatusBadRequest, "invalid device id")
+		api.ErrResponse(r.Context(), w, http.StatusBadRequest, "invalid device id")
 		return
 	}
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
-		api.ErrResponse(w, http.StatusInternalServerError, "failed to read request body")
+		api.ErrResponse(r.Context(), w, http.StatusInternalServerError, "failed to read request body")
```
therjak: do you want to log the error?

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -75,38 +75,37 @@ func getQueryParam(u *url.URL, param string) (string, error) {
...
 		return
 	}
 
 	var opts repository.KeyOptions
 	if err := json.Unmarshal([]byte(body), &opts); err != nil {
-		api.ErrResponse(w, http.StatusBadRequest, "invalid body")
+		api.ErrResponse(r.Context(), w, http.StatusBadRequest, "invalid body")
```
therjak: error log?

therjak (approved):

src/go/cmd/token-vendor/api/v1/v1.go
ensonic: Done.

src/go/cmd/token-vendor/api/v1/v1.go
ensonic: Also done. Thanks!

Issue #668, opened 2026-05-07
Migrate from nginx-ingress-controller to istio


ensonic:
Since the nginx-ingress controller is deprecated prepare the transition to istio using the gateways api through the followign steps:
* [x] add a config.sh variable USE_ISTIO (default false) and pass as a helm variable
* [x] add the istio resources to the codebase and use helm variable to install either nginx-ingress-controller or istio
* [x] port all the ingress resources in the codebase
* [x] verify tokenvendor-service as documented in
https://github.com/googlecloudrobotics/core/tree/main/src/go/cmd/token-vendor#curl-api-example-workflow

PR #670, opened 2026-05-07
Prepare istio resources

ensonic:
Add a script to generate the istio component manifests. Make use of the
new "use_istio" var on the base-cloud chart to activate either nginx or
istio.
    
Deployed to my personal testing project. This misses porting the
ingresses, but the chart at least comes up.

See #668


Tobias-Pe (approved):
LGTM, let's iterate over it on playground experiments.

PR #671, opened 2026-05-08
Use crypto/rand instead of math/rand

mp4096:
Why? `math/rand` is not supposed to be used
for sensitive applications, as its internal
state can be deduced by observing its outputs.
With the knowledge of the internal state, one
can predict the future outputs.

With `crypto/rand`, the outputs are harder to
predict at the cost of a slight performance hit:

Before:

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkCreateId-24    	 8559728	       136.0 ns/op
BenchmarkCreateId-24    	 8872668	       134.5 ns/op
BenchmarkCreateId-24    	 8619951	       133.9 ns/op
```

After:

```
goos: linux
goarch: amd64
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
BenchmarkCreateId-24    	 6881104	       172.6 ns/op
BenchmarkCreateId-24    	 6893991	       172.1 ns/op
BenchmarkCreateId-24    	 6859183	       173.3 ns/op
```


ensonic:
Please add a CL desc.

mp4096:
@ensonic done

ensonic (approved):
THanks!

PR #672, opened 2026-05-11
add token-vendor resources to helm chart values

koonpeng:
cherry pick from #665.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -1,5 +1,5 @@
 {
-  "lockFileVersion": 26,
+  "lockFileVersion": 18,
   "registryFileHashes": {
     "https://bcr.bazel.build/bazel_registry.json": "8a28e4aff06ee60aed2a8c281907fb8bcbf3b753c91fb5a5c57da3215d5b3497",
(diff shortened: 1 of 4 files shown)


MODULE.bazel.lock
drigz, line 1: I think you need to undo this? Reducing the lockFileVersion seems like a mistake.

koonpeng: Tried reverting it but doing a bazel build change it back again. Tried checking out main, make sure there are no changes, doing a bazel build still change the lock file version :thinking:.

koonpeng: I think https://github.com/googlecloudrobotics/core/pull/667 is made with a newer version of bazel that does not respect `.bazelversion`.

drigz: Hmm, I'm worried about merging the change given https://github.com/googlecloudrobotics/core/pull/667/changes suggests we might break the build elsewhere. Maybe you can just split it out of this commit locally and ignore the fact that Bazel is trying to undo it until @ensonic can let us know if any changes are necessary in the workspace for #667?

koonpeng: done

drigz (approved):

PR #673, opened 2026-05-11
Port token-vendor ingress to http routes.

ensonic:
Move the configuration of the istio-loadbalancer to the gateway. Now
istio is generating the loadbalancer to match the gateway settings.
This lets us simplify the configuration when updating istio in
third_party.

See #668


Tobias-Pe:
How will we add the gateway resource ?
Or is this added by default istio ? Or do u add it urself in your dev env for now

Tobias-Pe (approved):
Marvelous!!!

ensonic:
> How will we add the gateway resource ? Or is this added by default istio ? Or do u add it urself in your dev env for now

The gateway resource is always manaully added. Its in istio.yaml

PR #674, opened 2026-05-11
Add missing istio.yaml

ensonic:
This creates the gateway and was part of previous change, but somehow missign from the PR.

See #668

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -1,0 +1,41 @@
+{{- if eq .Values.use_istio "true" }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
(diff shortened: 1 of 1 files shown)


src/app_charts/base/cloud/istio.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if eq .Values.use_istio "true" }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: gateway
```
Tobias-Pe: There could be use cases for multiple gateways, therefore id give this a more specific name e.g. cloud-robotics-gateway etc.

We can leave it also for now and rethink further on once we get past derisking

Tobias-Pe (approved):

PR #678, opened 2026-05-13
disable grafana test framework so postsubmit does not wait for it to be ready

koonpeng:
The test cluster is updated recently and it does not include `prom-grafana-test`, it causes postsubmit CI to fail as it tries to wait for the pod to be ready.


koonpeng:
Manual postsubmit passes https://github.com/googlecloudrobotics/core/actions/runs/25774754012

ensonic (approved):
I saw this on prod as well. Thanks for figuring out how to turn it off.

PR #679, opened 2026-05-13
fail app_chart build when there is lint error

koonpeng:
Tested by removing https://github.com/googlecloudrobotics/core/blob/fc4553d7b8243cbe91a2b125c7434ba76e2bf26c/src/app_charts/k8s-relay/values-robot.yaml#L6-L7. Build fails with a lint error.


ensonic (approved):
Nice! Thanks!

PR #680, opened 2026-05-15
Switch to PathPrefix for the token-vendor HTTPRoute

ensonic:
This is simpler than the regexp which we don't need here.

### DIFF
--- a/src/app_charts/token-vendor/cloud/http-route.yaml
+++ b/src/app_charts/token-vendor/cloud/http-route.yaml
@@ -17,8 +17,8 @@ spec:
       port: 80
     matches:
     - path:
-        type: RegularExpression
-        value: (?i)/apis/core.token-vendor/v1/public-key.read.*
+        type: PathPrefix
+        value: /apis/core.token-vendor/v1/public-key.read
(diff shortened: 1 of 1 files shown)


src/app_charts/token-vendor/cloud/http-route.yaml
```diff
@@ -17,8 +17,8 @@ spec:
       port: 80
     matches:
     - path:
-        type: RegularExpression
-        value: (?i)/apis/core.token-vendor/v1/public-key.read.*
```
csieber, line 21: Any idea why the `(?i)` was used here? A prefix match would be case-sensitive, not case-insensitive

ensonic: Probably some i2gw weirdness. This is the ingress defs:
https://github.com/googlecloudrobotics/core/blob/main/src/app_charts/token-vendor/cloud/ingress.yaml

and the docs also say this is case-sensitive: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types

csieber (approved):

PR #683, opened 2026-05-18
Update the gcloud auth commands.

ensonic:
We need the `application-default` variant.


src/bootstrap/cloud/terraform/service-account.tf
```diff
@@ -106,7 +106,7 @@ resource "google_service_account_iam_member" "human-acl-act-as-self" {
 
 # Grant permissions to generate tokens for registering new workcell clusters.
 # This lets users run:
-#   gcloud auth print-access-token \
+#   gcloud auth application-default print-access-token \
```
drigz, line 109: FYI This command here is good if you want a token that will fail the final auth check, just use a different project ID in this command than the token-vendor you're talking to.

drigz (approved):

PR #684, opened 2026-05-18
Add gRPC integration tests for http relay and update docs

dfreese:
This adds an integration test that demonstrates the use of different
gRPC types over the http relay (unary, server-streaming,
client-streaming, and bi-directional streaming).  I have updated the
documentation to reflect that this works.  I was never able to reproduce
the specific failure mentioned in the docs.  From my other experiments,
server-streaming was working without issue, so I think this integration
test is representative of the current state.  I haven't attempted to
trace if this was fixed by a particular commit.


src/go/tests/relay/grpc_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+	relayAddr := fmt.Sprintf("localhost:%d", relayEnv.relayPort)
+	conn, err := grpc.Dial(relayAddr, grpc.WithInsecure())
+	if err != nil {
+		t.Fatalf("Failed to create client connection: %v", err)
+	}
+	defer conn.Close()
+
+	grpcClient := pb.NewTestServiceClient(conn)
+	ctx := metadata.AppendToOutgoingContext(context.Background(), "x-server-name", "server_name")
```
drigz: t.Context() will be cancelled when the test ends so saves the extra defer. You probably still want a timeout as the default timeout is 10 minutes.

src/go/tests/relay/grpc_test.go
```diff
@@ -0,0 +1,349 @@
...
+	t.Logf("Received: %q", resp.Value)
+	if want := "Echo: " + reqVal; resp.Value != want {
+		t.Errorf("got %q, want %q", resp.Value, want)
+	}
+}
+
+func TestServerStreamingOverRelay(t *testing.T) {
+	relayEnv := mustInitRelay(t)
+
+	lis, err := net.Listen("tcp", fmt.Sprintf("localhost:%d", relayEnv.backendPort))
```
drigz, line 187: It'd be cleaner to lis.Close() even though leaking a few ports until end of test process is harmless.

WDYT about a helper:

```
grpcClient, cleanup := relayEnv.mustCreateClient(t)
defer cleanup()
```

drigz (approved):
This is really nice, thank you so much!

src/go/tests/relay/grpc_test.go
```diff
@@ -0,0 +1,349 @@
...
+	defer conn.Close()
+
+	grpcClient := pb.NewTestServiceClient(conn)
+	ctx := metadata.AppendToOutgoingContext(context.Background(), "x-server-name", "server_name")
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	reqVal := "hello"
+	t.Logf("Calling ServerStream with %q", reqVal)
+	stream, err := grpcClient.ServerStream(ctx, &pb.StringMessage{Value: reqVal})
```
ensonic, line 214: I'd use a local var e.g. `numPackets := 5`, pass it here as a param and also use in the lop below to make sure this matches.

src/go/tests/relay/grpc_test.go
```diff
@@ -0,0 +1,349 @@
...
+		config.ForceHttp2 = true // Enable HTTP/2 for gRPC
+		config.NumPendingRequests = 5
+		relayClient := client.NewClient(config)
+		relayClient.Start()
+	}()
+
+	relayHealthy := false
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		relayHealthzAddr := fmt.Sprint("http://localhost:", relayPort, "/healthz")
```
ensonic: nit-1: I'd move this out of the loop, maybe right after you pick the ports.
nit-2: for consistency (I first though the format args are misisng)
relayHealthzAddr := fmt.Sprintf("http://localhost:%d/healthz", relayPort)

src/go/tests/relay/grpc_test.go
```diff
@@ -0,0 +1,349 @@
...
+	pb.RegisterTestServiceServer(grpcServer, &testServer{})
+	go func() {
+		if err := grpcServer.Serve(lis); err != nil {
+			t.Fatalf("grpcServer.Serve(lis) failed unexpectedly: %v", err)
+		}
+	}()
+	defer grpcServer.Stop()
+
+	relayAddr := fmt.Sprintf("localhost:%d", relayEnv.relayPort)
+	conn, err := grpc.Dial(relayAddr, grpc.WithInsecure())
```
ensonic: just inline like above (the var is not used elsewhere), also in the other tests.
conn, err := grpc.Dial(fmt.Sprintf("localhost:%d", relayEnv.relayPort), grpc.WithInsecure())

ensonic (approved):
Thanks a lot!

src/go/tests/relay/grpc_test.go
dfreese: Done.

src/go/tests/relay/grpc_test.go
dfreese: Done

src/go/tests/relay/grpc_test.go
dfreese: Added a constant at the top.

src/go/tests/relay/grpc_test.go
dfreese: Swapped to t.Context(), though I left the defer because it seems better to be thorough remind folks that cancel should always be called.

src/go/tests/relay/grpc_test.go
dfreese: This is passed into `grpcServer.Serve(lis)` and [`Serve`](https://pkg.go.dev/google.golang.org/grpc#Server.Serve) states that:

> Serve returns when lis.Accept fails with fatal errors. lis will be closed when this method returns.

I'm going to leave this as is.  Let me know if I missed something.

PR #685, opened 2026-05-19
Move sync namespace check to helper.

ensonic:
This will let us add extra namespaces as needed.
See PR #676

### DIFF
--- a/src/go/pkg/synk/synk.go
+++ b/src/go/pkg/synk/synk.go
@@ -23,6 +23,7 @@ import (
 	"log/slog"
 	"reflect"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


src/go/pkg/synk/synk.go
```diff
@@ -410,6 +411,15 @@ func (s *Synk) applyAll(
 	return results, err
 }
 
+func validateNamespace(r *unstructured.Unstructured, optsNs string) error {
+	ns := r.GetNamespace()
+	allowd := []string{"kube-system", optsNs}
+	if ns == "" || !slices.Contains(allowd, ns) {
```
Tobias-Pe: The previous code had  ns != "" &&... --> Throw error 

This here is the inversion of it. 

So if my logic is not twisted previous logic was: If a namespace is provided, it must be either the expected namespace (opts.Namespace) or kube-system. If no namespace is provided, that's fine too.

This is a change in the logic. Just raising this if not intended

ensonic: Thanks for carefully reading. Restored the logic. PTAL

ensonic:
@awhdesmond FYI. In case you need to modify the token-vendor chart in a way to e.g. deploy to default ns.

Tobias-Pe (approved):
Thanks!

PR #687, opened 2026-05-20
Add istio meshConfig, internal auth gateway and token-vendor auth httproutes

awhdesmond:
1. Configure extension providers in Istio's mesh_config to use `crc-auth-gateway`
2. Deploy `crc-auth-gateway` using Gateway API `Gateway` resource, but only give it an internal clusterIP.
3. Apply `AuthorizationPolicy` on `crc-gateway` to use `crc-auth-gateway` extension provider for external HTTP authorization.
4. Add auth HTTPRoute for `public-key-access-auth` and `public-key-access-auth`, targetting `crc-auth-gateway`
5. Include an explicity-allow `catch-all` route for `crc-auth-gateway` as not all paths require authz. (It can be switched to explicit-deny)

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -121,7 +121,7 @@ spec:
     spec:
       containers:
       - name: nginx
-        image: nginxinc/nginx-unprivileged:alpine
+        image: nginxinc/nginx-unprivileged:1.30.1-alpine
         ports:
         - containerPort: 8080
         command: ["/bin/sh", "-c"]
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -5,7 +5,7 @@ metadata:
   name: crc-gateway
   annotations:
     # Use standard K8s annotations for cloud-specific LB settings
-    networking.istio.io/service-type: LoadBalancer 
+    networking.istio.io/service-type: LoadBalancer
     networking.istio.io/service-externalTrafficPolicy: Local
 spec:
   gatewayClassName: istio # Provisions Istio-managed LB infrastructure
@@ -50,18 +50,18 @@ spec:
 ---
 apiVersion: gateway.networking.k8s.io/v1
 kind: Gateway
-metadata: 
+metadata:
   name: crc-auth-gateway
   annotations:
     networking.istio.io/service-type: ClusterIP # only require internal cluster ip
-spec: 
-  gatewayClassName: istio # Provisions Istio-managed LB 
+spec:
+  gatewayClassName: istio # Provisions Istio-managed LB
   listeners:
     - name: http
       hostname: {{ .Values.domain }}
       protocol: HTTP
       port: 8080
-      allowedRoutes: 
+      allowedRoutes:
         namespaces:
           from: All
 ---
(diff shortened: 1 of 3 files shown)


src/app_charts/base/cloud/istio.yaml
```diff
@@ -47,4 +47,103 @@ spec:
...
+    matchLabels:
+      app: explicit-allow-svc
+  template:
+    metadata:
+      labels:
+        app: explicit-allow-svc
+    spec:
+      containers:
+      - name: nginx
+        image: nginxinc/nginx-unprivileged:alpine
```
awhdesmond: Will we be allowed to use [open-source (Apache 2 license) NGINX image](https://github.com/nginx/docker-nginx-unprivileged) here for CRC?

src/app_charts/base/cloud/istio.yaml
```diff
@@ -47,4 +47,103 @@ spec:
...
+        value: /
+    backendRefs:
+    - name: explicit-allow-svc
+      namespace: default
+      port: 8080
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: explicit-allow-svc
```
awhdesmond, line 109: It's may be worth to refactor this service out to its own file.

koonpeng: Yeah, it appears to be only for testing?

src/app_charts/base/cloud/istio.yaml
koonpeng: Not authoritative, but CRC itself is apache-2.0 so I would assume yes.

src/app_charts/base/cloud/istio.yaml
awhdesmond: This service is used as the catch-all service with pathprefix `/` under the auth gateway. I would assume not all paths require authz, so this catch-all service will respond with HTTP 200 for all those services.

src/app_charts/base/cloud/istio.yaml
awhdesmond: i have updated to use `nginxinc/nginx-unprivileged:1.30.1-alpine`

ensonic: Yes, the license is fine. We can also consider a simple go-lang server if all it needs to do is to reply with OK 200. But that can be a followup.

src/app_charts/base/cloud/istio.yaml
ensonic: Yes, we need to support routes that don't require auth.

ensonic: Can we try this:
```
filters:
    - type: RequestRedirect
      requestRedirect:
        statusCode: 200
```

awhdesmond: will result in:
``` 
The HTTPRoute "crc-auth-catchall" is invalid: 
* spec.rules[0].filters[0].requestRedirect.statusCode: Unsupported value: 200: supported values: "301", "302"
* <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

ensonic (approved):
Awesome, lets iterate in followups to finetune it.

PR #688, opened 2026-05-20
Use helm templating to hide the HTTPRoute duplication

ensonic:
Use a range loop to produce both the traffic and the auth routes in-place. This will avoid having to specify all path matches twice.

```shell
$ kubectl get httproute -n app-token-vendor
NAME                      HOSTNAMES                                    AGE
public-key-access         ["www.endpoints.XXX.cloud.goog"]   9d
public-key-access-auth    ["www.endpoints.XXX.cloud.goog"]   6h6m
public-key-manager        ["www.endpoints.XXX.cloud.goog"]   9d
public-key-manager-auth   ["www.endpoints.XXX.cloud.goog"]   6h6m
token-vendor              ["www.endpoints.XXX.cloud.goog"]   9d
```


csieber (approved):
good idea!

PR #689, opened 2026-05-21
Port remaining ingress resources to http routes

ensonic:
Instead of `rewrite-target` use ReplacePrefixMatch.

See #668

### DIFF
--- a/src/app_charts/base/cloud/kubernetes-api.yaml
+++ b/src/app_charts/base/cloud/kubernetes-api.yaml
@@ -1,4 +1,4 @@
-{{ if eq .Values.onprem_federation "true" }}
+{{- if eq .Values.onprem_federation "true" }}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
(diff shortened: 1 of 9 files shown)


src/app_charts/base/cloud/kubernetes-api.yaml
```diff
@@ -23,4 +23,46 @@ spec:
...
+        path:
+          type: ReplacePrefixMatch
+          replacePrefixMatch: /
+    backendRefs:
+    - name: kubernetes
+      port: 443
+{{- end }}
+---
+{{- end }}
+{{- end }}
```
Tobias-Pe, line 68: Why the double end?

awhdesmond (approved):

src/app_charts/base/cloud/kubernetes-api.yaml
ensonic: first one is from:
{{- range list "" "-auth" }}
second one is from
{{- if eq .Values.onprem_federation "true" }}

PR #691, opened 2026-05-27
✨ istio: Inject Gateway API CRDs standard-install into config generation

Tobias-Pe:
This change integrates the download of standard Gateway API CRDs directly into the Istio config generation script.

#### Why

Gateway API CRDs like GRPCRoute are nice to have and we want them added to support gRPC traffic routing. Since the GKE-managed Gateway API standard channel does not deploy GRPCRoute automatically, bundling this CRD into Istio's generated templates ensures they are present in GKE clusters when Istio is active.

- Add GATEWAY_API_VERSION variable at the top of update-istio.sh.
- Modify Step 3 in update-istio.sh to download and prepend the grpcRoute CRD manifest to the generated Istio template.

```
❯ kubectl --context=gke_robco-ensonic_europe-west1-c_cloud-robotics get crd grpcroutes.gateway.networking.k8s.io
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "grpcroutes.gateway.networking.k8s.io" not found

> deploy.sh update robco-ensonic
...
[SUCCESS] Update apiextensions.k8s.io/v1/CustomResourceDefinition /gateways.networking.istio.io
[SUCCESS] Update apiextensions.k8s.io/v1/CustomResourceDefinition /grpcroutes.gateway.networking.k8s.io
...

❯ kubectl --context=gke_robco-ensonic_europe-west1-c_cloud-robotics get crd grpcroutes.gateway.networking.k8s.io
NAME                                   CREATED AT
grpcroutes.gateway.networking.k8s.io   2026-05-27T10:14:11Z
```

TESTED=on playground project & deployed dummy grpcroute


Tobias-Pe:
I think this is the better alternative to #690

third_party/istio/update-istio.sh
```diff
@@ -55,9 +57,19 @@ cat ${tmpdir}/istio_full.yaml \
 dst="${SCRIPT_DIR}/istio-generated.yaml"
 {
   echo '{{- if eq .Values.use_istio "true" }}'
-  echo "# Generated by update-istio.sh."
+  echo "# Generated by update-istio.sh. DO NOT EDIT!"
+  echo "# ---------------------------------------------------------"
+  echo "# Gateway API CRDs (Standard Channel v${GATEWAY_API_VERSION})"
+  echo "# See: https://docs.cloud.google.com/service-mesh/v1.27/docs/gateway/prepare-gateway#install_custom_resource_definitions"
+  echo "# ---------------------------------------------------------"
+  curl -fsSL "https://github.com/kubernetes-sigs/gateway-api/releases/download/v${GATEWAY_API_VERSION}/standard-install.yaml"
```
drigz: WDYT about just adding GRPCRoute and linking to https://github.com/googlecloudrobotics/core/blame/efa7dfa8f25682ebfdca3543b96e6e1c10e56a42/src/bootstrap/cloud/terraform/cluster.tf#L41 to say that that installs the rest of the Gateway API CRDs?

Otherwise I'm not sure if we'll end up with the versions chosen by GKE or the versions chosen by us.

third_party/istio/update-istio.sh
```diff
@@ -55,9 +57,19 @@ cat ${tmpdir}/istio_full.yaml \
...
+  echo "---"
+  echo "# ---------------------------------------------------------"
+  echo "# Istio System Manifests"
```
drigz, line 69: Do we need to change the Istio values to also apply the GatewayClass that lets you say your Gateway should be provided by Istio?

drigz (approved):
Approving as right now this only affects the test projects already using Istio, and we can figure out async what the right way to manage CRD creation/update is.

third_party/istio/update-istio.sh
Tobias-Pe: https://docs.cloud.google.com/service-mesh/v1.27/docs/gateway/prepare-gateway#install_custom_resource_definitions just added the whole thing so i aligned with that.

But yes SGTM

src/bootstrap/cloud/terraform/.terraform.lock.hcl
```diff
@@ -2,20 +2,20 @@
...
-    "zh:2c75a1ea53b21646681e49fbdf0a599817c2f400f1e73d7779f0e3e1d230e6f3",
-    "zh:34ab9dab67230adaee6a9cd6861cba969555777ca6eb0ae1d2ac7b1f3cb73832",
-    "zh:45d5d7ee38fb7bf58dd19b774dd637f3cb9caef1d1930dde594467dde7fdea50",
-    "zh:651ffd36697d8268471d50d0fae664549b7f1e627c03d6e90f80172947f7b1d4",
-    "zh:8557db0beb201ba8ba70a7a38ba8d1ce9ffb9e98c616b89f6e3c2203e0528803",
-    "zh:b6a2e53809e0827cb7c47b1279c3511223898f7e3c1536f74ba057d99c72c2e9",
-    "zh:bf4aea9d1eb663df9d458c974d2e3f9ca7f724280a103706d0a2b1597593c7af",
-    "zh:c5f0160e0658d75b4a339b18f7a544e721f3750e26b97fc4887e98b8e085ffff",
-    "zh:e41a215491c64b535eda5585139dab0a32836e631b470cc520ceef4aa9ce7748",
-    "zh:e490061ab15c6053c651a7996a9c77cacdb83528ed49d0460c9621e93ad7d0cf",
```
Tobias-Pe, line 18: hm i didn't intentionally change this, do you want this out or in ?

third_party/istio/update-istio.sh
Tobias-Pe: I think i do not understand what you are aiming at 

We are providing gateway classes when use_istio is on. Those need to be referenced by your application routes, to get traffic routed to your pods.

src/bootstrap/cloud/terraform/.terraform.lock.hcl
drigz: probably better out or in a separate PR if it breaks something

third_party/istio/update-istio.sh
drigz: aha, of course, that's already there. sorry for the noise!

src/bootstrap/cloud/terraform/.terraform.lock.hcl
Tobias-Pe: Okidoki. Done 🧹

PR #692, opened 2026-05-28
Upgrade NGINX version to remediate CVE-2026-9256

awhdesmond:
Upgrade NGINX version to remediate [CVE-2026-9256](https://www.cve.org/CVERecord?id=CVE-2026-9256)

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -121,7 +121,7 @@ spec:
     spec:
       containers:
       - name: nginx
-        image: nginxinc/nginx-unprivileged:1.30.1-alpine
+        image: nginxinc/nginx-unprivileged:1.31.1-alpine
         ports:
         - containerPort: 8080
         command: ["/bin/sh", "-c"]


src/app_charts/base/cloud/istio.yaml
```diff
@@ -121,7 +121,7 @@ spec:
     spec:
       containers:
       - name: nginx
-        image: nginxinc/nginx-unprivileged:1.30.1-alpine
+        image: nginxinc/nginx-unprivileged:1.31.1-alpine
```
Tobias-Pe, line 124: 1.31 seems to be mainline and not stable. I don't like the implications of switching away from stable for prod systems. 

1.30.2 should have the fix in and is still the version for stable.

https://my.f5.com/manage/s/article/K000161377#:~:text=1.31.0%0A1.0.0%20%2D%201.30.1-,1.31.1%0A1.30.2,-High/8.1%20(CVSS

IMO we should switch the tag to nginxinc/nginx-unprivileged:1.30-alpine, then  we get the 1.30.X upgrades automatically. WDYT @awhdesmond @csieber  ?

Tobias-Pe (changes requested):

src/app_charts/base/cloud/istio.yaml
Tobias-Pe: Proof of 1.30-alpine working 
```bash
docker run --rm nginxinc/nginx-unprivileged:1.30-alpine nginx -v

Unable to find image 'nginxinc/nginx-unprivileged:1.30-alpine' locally
1.30-alpine: Pulling from nginxinc/nginx-unprivileged
Digest: sha256:df0e9edf92b8436ff797fe5a2cbfc66be1df775c113d322ccadf5c7f3100eda8
Status: Downloaded newer image for nginxinc/nginx-unprivileged:1.30-alpine
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perfo
rm configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-defau
lt.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d
/default.conf
10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs fr
om the packaged version
/docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.s
h
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.s
h
/docker-entrypoint.sh: Configuration complete; ready for start up
nginx version: nginx/1.30.2
```

Tobias-Pe:
```suggestion
        image: nginxinc/nginx-unprivileged:1.30-alpine
```

Tobias-Pe: Otherwise there is also stable-alpine but i think automatic major/minor upgrades can come risky and if something fails it's super unclear. 

Automatic Patches OTOH should be best practice

csieber: I would usually prefer we pin by hash and have automation around updating the hash. However it seems we don't have this here yet. 
I am fine with both options. If you have time to monitor the rollout closely we can do `1.30.1` -> `1.31.1`. Otherwise we can do `1.30` which has less risk to change behavior.

awhdesmond: Usually, I believe mainline is [recommended for production](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/) but in our case, where it's more difficult to release to production, I would agree that going back to stable `1.30.2-alpine` is a more appropriate option! 

However, I would prefer to not use floating tags such as `1.30-alpine`, even though we get automatic 1.30.X upgrades. Using immutable tags `1.30.2-alpine` increases the visibility & auditability of the change. It also easier to rollback if the patch version broke something (albeit unlikely). Ideally, in GitOps, we can bump the patch version and there will be a diff, and ArgoCD and sync the updated manifest and rollout the deployment to the new image for us. If we use floating tags, ArgoCD can't detect the change, and we need to set `imagePullPolicy: Always` to pull the new image when restarting. 

If we want to achieve automatic patching via floating tags, I believe the suggesting by @csieber where we pin by hash and have automation using something a tool like dependabot to update the hash would be more ideal.

WDYT?

Tobias-Pe: > mainline is [recommended for production](https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/)

Good finding ! Hm then yes, in that case, jumping to mainline i will not block. We havn't anyways started using istio, so lets start it off with the right versioning base 👍 

>  not use floating tags such as 1.30-alpine

Ah okay, my argo / gitops knowledge is not very profound (learning as i use currently). Good to know :)

Tobias-Pe (approved):

PR #693, opened 2026-06-02
Invert catch-all behavior for auth gateway

awhdesmond:
Invert catch-all behavior in auth gateway to return HTTP 403 Forbidden by default.

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -131,7 +131,7 @@ spec:
               listen 8080;
               location / {
                   add_header Content-Type text/plain;
-                  return 403 "Forbidden\n";
+                  return 404 "Not Found\n";
               }
           }' > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'
 ---
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -99,25 +99,25 @@ spec:
         type: PathPrefix
         value: /
     backendRefs:
-    - name: explicit-allow-svc
+    - name: crc-auth-gateway-catchall-svc
       namespace: default
       port: 8080
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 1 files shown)


csieber (approved):
thanks!

Tobias-Pe (approved):
Thanks!

Tobias-Pe (changes requested):
So which code should we return bc as you stated @awhdesmond the 403 will create a signin redirect. 

404 doesn't seem as a bad option (first thought)

awhdesmond:
> So which code should we return bc as you stated @awhdesmond the 403 will create a signin redirect.
> 
> 404 doesn't seem as a bad option (first thought)

What about catchall service return an unsued HTTP 4xx, gets intercepted by EnvoyFilter which then converts it back to HTTP 403?

csieber:
> What about catchall service return an unsued HTTP 4xx, gets intercepted by EnvoyFilter which then converts it back to HTTP 403?

I am in favor of the 404 because that feels like something a server should return if the path is not registered.

Tobias-Pe:
> > What about catchall service return an unsued HTTP 4xx, gets intercepted by EnvoyFilter which then converts it back to HTTP 403?
> 
> I am in favor of the 404 because that feels like something a server should return if the path is not registered.

This would also help keep things straightforward. 

The envoyfilter is a great option and very powerful but not as clear when it comes to trying to follow what is happening. So we should keep our filter usage minimal

src/app_charts/base/cloud/istio.yaml
```diff
@@ -131,19 +131,19 @@ spec:
               listen 8080;
               location / {
                   add_header Content-Type text/plain;
-                  return 200 "OK\n";
+                  return 403 "Forbidden\n";
```
Tobias-Pe:
```suggestion
                  return 404 "Not Found\n";
```

awhdesmond: Updated

Tobias-Pe (approved):
Thank you Desmond !

PR #695, opened 2026-06-04
HTTP status echo server for auth gateway catchall

awhdesmond:
Simple HTTP server with configurable response HTTP status code. The aim is to replace NGINX with this as the catchall service for internal auth gateway.


ensonic (approved):
Thanks!

PR #700, opened 2026-06-10
⏪️ Revert "🛂 istio: whitelist cookie header in crc-auth-gateway provider"

Tobias-Pe:
Reverts googlecloudrobotics/core#699

This is intrinsic specific and not fitting for the opensource community. I will add this as a post processing to the deployment scripts on intrinsic side


ensonic (approved):
I think the cookie header is not an issues. It is an official header.

Tobias-Pe:
Uhhm, revert revert then ? haha

PR #701, opened 2026-06-11
Replace auth-gateway catchall NGINX using HTTP status echo server

awhdesmond:
Replace auth-gateway catchall NGINX using the newly implemented HTTP status echo server.

### DIFF
--- a/src/app_charts/base/BUILD.bazel
+++ b/src/app_charts/base/BUILD.bazel
@@ -108,6 +108,7 @@ app_chart(
         "app-rollout-controller": "//src/go/cmd/app-rollout-controller:app-rollout-controller-image",
         "chart-assignment-controller": "//src/go/cmd/chart-assignment-controller:chart-assignment-controller-image",
         "cr-syncer-auth-webhook": "//src/go/cmd/cr-syncer-auth-webhook:cr-syncer-auth-webhook-image",
+        "status-echo-server": "//src/go/cmd/status-echo-server:status-echo-server-image",
     },
     values = "values-cloud.yaml",
     visibility = ["//visibility:public"],
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -120,20 +120,19 @@ spec:
         app: crc-auth-gateway-catchall-svc
     spec:
       containers:
-      - name: nginx
-        image: nginxinc/nginx-unprivileged:1.31.1-alpine
+      - name: catchall
+        image: {{ .Values.registry }}{{ .Values.images.status_echo_server }}
         ports:
         - containerPort: 8080
-        command: ["/bin/sh", "-c"]
         args:
-        - |
-          echo 'server {
-              listen 8080;
-              location / {
-                  add_header Content-Type text/plain;
-                  return 404 "Not Found\n";
-              }
-          }' > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'
+        - --port=8080
+        - --http_status=404
+        readinessProbe:
+          tcpSocket:
+            port: 8080
+        livenessProbe:
+          tcpSocket:
+            port: 8080
 ---
 apiVersion: v1
 kind: Service
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -123,7 +123,9 @@ spec:
       - name: catchall
         image: {{ .Values.registry }}{{ .Values.images.status_echo_server }}
         ports:
-        - containerPort: 8080
+        - name: http
+          containerPort: 8080
+          protocol: TCP
         args:
         - --port=8080
         - --http_status=404
@@ -133,6 +135,19 @@ spec:
         livenessProbe:
           tcpSocket:
             port: 8080
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          privileged: false
+          capabilities:
+              drop:
+              - ALL
+          seccompProfile:
+            type: RuntimeDefault
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65532
+        runAsGroup: 65532
 ---
 apiVersion: v1
 kind: Service


awhdesmond:
Please, let me know if I am missing anything w.r.t to using the status-echo-server image in CRC helm chart! Thanks!

Tobias-Pe (commented):
We need also an allow service I think right? For services that do not require auth e.g. no aurh-url annotation

Is there a possibility to define the returned status code in the route rather than the deployment args?

awhdesmond:
> We need also an allow service I think right? For services that do not require auth e.g. no aurh-url annotation
> 
> Is there a possibility to define the returned status code in the route rather than the deployment args?

This catch-all service will handle all paths that do not have `auth-url` annotation. I recall our earlier discussion was to return HTTP 404 for those.

Tobias-Pe:
> > We need also an allow service I think right? For services that do not require auth e.g. no aurh-url annotation
> > Is there a possibility to define the returned status code in the route rather than the deployment args?
> 
> This catch-all service will handle all paths that do not have `auth-url` annotation. I recall our earlier discussion was to return HTTP 404 for those.

The catch-all is supposed to handle all services that are not correctly or not at all configured in the auth gateway. Its a general security extra step, so that if someone without knowledge creates routes, that the service is not immediately out in the internet

Services with auth-url in the migration need to get a route pointing to e.g. the auth-proxy.

Services without auth-url in the migration will have to get a route in the auth gateway that returns 200, so it doesn't get caught by the catch-all.

awhdesmond:
> Services without auth-url in the migration will have to get a route in the auth gateway that returns 200, so it doesn't get caught by the catch-all.

I understand. Currently there is no way to perform direct response on the HTTPRoute itself. However, it's possible to deploy an addition `status-echo-server` that returns HTTP 200 and set it as the backendRef for those paths that originally did not have `auth-url`

Tobias-Pe:
> > Services without auth-url in the migration will have to get a route in the auth gateway that returns 200, so it doesn't get caught by the catch-all.
> 
> I understand. Currently there is no way to perform direct response on the HTTPRoute itself. However, it's possible to deploy an addition `status-echo-server` that returns HTTP 200 and set it as the backendRef for those paths that originally did not have `auth-url`

Sorry, i wasn't in the loop when `status-echo-server` got designed: Could we modify the echo server to echo always depending on e.g. query param. Then we could leverage the same deployment for catch-all-deny and no-auth-allow routes. And have this all in the routes defined. I'd imagine routes like these: 

404 deny all:
```yaml
spec:
  hostnames:
  - blabla
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: crc-auth-gateway
    namespace: default
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: status-echo-server
      namespace: default
      port: 8080
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplaceFullPath
              replaceFullPath: /echo?status=404
```

200 service allow:
```yaml
spec:
  hostnames:
  - blabla
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: crc-auth-gateway
    namespace: default
    sectionName: http
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: status-echo-server
      namespace: default
      port: 8080
      weight: 1
    - matches:
      - path:
          type: PathPrefix
          value: /my_favorite_unauthenticated_super_secure_service/
      filters:
        - type: URLRewrite
          urlRewrite:
            path:
              type: ReplaceFullPath
              replaceFullPath: /echo?status=200
```

This is how we would do it right now for to be authenticated services at Intrinsic with the URLRewrite

Tobias-Pe:
> > > Services without auth-url in the migration will have to get a route in the auth gateway that returns 200, so it doesn't get caught by the catch-all.
> > 
> > 
> > I understand. Currently there is no way to perform direct response on the HTTPRoute itself. However, it's possible to deploy an addition `status-echo-server` that returns HTTP 200 and set it as the backendRef for those paths that originally did not have `auth-url`
> 
> Sorry, i wasn't in the loop when `status-echo-server` got designed: Could we modify the echo server to echo always depending on e.g. query param. Then we could leverage the same deployment for catch-all-deny and no-auth-allow routes. And have this all in the routes defined. I'd imagine routes like these:
> 
> 404 deny all:
> 
> ```yaml
> spec:
>   hostnames:
>   - blabla
>   parentRefs:
>   - group: gateway.networking.k8s.io
>     kind: Gateway
>     name: crc-auth-gateway
>     namespace: default
>     sectionName: http
>   rules:
>   - backendRefs:
>     - group: ""
>       kind: Service
>       name: status-echo-server
>       namespace: default
>       port: 8080
>       weight: 1
>     matches:
>     - path:
>         type: PathPrefix
>         value: /
>       filters:
>         - type: URLRewrite
>           urlRewrite:
>             path:
>               type: ReplaceFullPath
>               replaceFullPath: /echo?status=404
> ```
> 
> 200 service allow:
> 
> ```yaml
> spec:
>   hostnames:
>   - blabla
>   parentRefs:
>   - group: gateway.networking.k8s.io
>     kind: Gateway
>     name: crc-auth-gateway
>     namespace: default
>     sectionName: http
>   rules:
>   - backendRefs:
>     - group: ""
>       kind: Service
>       name: status-echo-server
>       namespace: default
>       port: 8080
>       weight: 1
>     - matches:
>       - path:
>           type: PathPrefix
>           value: /my_favorite_unauthenticated_super_secure_service/
>       filters:
>         - type: URLRewrite
>           urlRewrite:
>             path:
>               type: ReplaceFullPath
>               replaceFullPath: /echo?status=200
> ```
> 
> This is how we would do it right now for to be authenticated services at Intrinsic with the URLRewrite

Gemini just told me 

>URLRewrite Query Parameters (High Impact): Almost all companion auth routes use ReplaceFullPath to rewrite the path to /auth-proxy/authorize?scopes=apikey (carrying over the query parameters from the original auth-url annotation). Standard Gateway API does not support query parameters in ReplaceFullPath filters. This will cause validation failures on standard controllers and must be resolved manually (e.g. by using custom filters or moving scope configuration to headers).

Let me double check this if this is true :O

awhdesmond:
> Let me double check this if this is true :O

If query parameters present a problem, why not something like `/status/200` and `/status/404` ?

Tobias-Pe:
> > Let me double check this if this is true :O
> 
> If query parameters present a problem, why not something like `/status/200` and `/status/404` ?

Yes an option for status echo, but for Intrinsic auth-proxy this could be a risky rewrite :O 
But this topic / risk is unrelated to cloudrobotics: i'll open a chat thread about it once i have investigated for intrinsic

ensonic (approved):
+1 for replacing nginx. Maybe we can resolve this iteratively (that is discuss missing features maybe in a separate ticket).

PR #702, opened 2026-06-11
Bump k8s client libs.

ensonic:
* adjust for controller runtime api-changes (introduced generics).
* add a common wrt client-go versions and test issues in never versions.

This should be resolving the failed dependabot PR we're repeatedly got.


src/go/pkg/controller/approllout/controller.go
```diff
@@ -80,81 +82,80 @@ func Add(ctx context.Context, mgr manager.Manager, baseValues chartutil.Values)
...
-				change = change || e.ObjectOld.GetName() != e.ObjectNew.GetName()
-				if change {
-					slog.Info("AppRollout controller received update event", slog.String("Robot", e.ObjectNew.GetName()))
-					r.enqueueAll(ctx, q)
-				}
-			},
-			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.RateLimitingInterface) {
-				slog.Info("AppRollout controller received delete event", slog.String("Robot", e.Object.GetName()))
-				time.AfterFunc(3*time.Second, func() {
+		source.Kind(mgr.GetCache(), kclient.Object(&registry.Robot{}),
```
csieber, line 113: The diff here is hard to view, did some formatting change?

src/go.mod
```diff
@@ -24,29 +22,31 @@ require (
...
-	google.golang.org/protobuf v1.36.11
+	google.golang.org/protobuf v1.36.12-0.20260120151049-f2248ac996af
 	gopkg.in/h2non/gock.v1 v1.1.2
-	k8s.io/api v0.28.4
-	k8s.io/apiextensions-apiserver v0.28.4
-	k8s.io/apimachinery v0.28.4
-	k8s.io/cli-runtime v0.28.4
-	k8s.io/client-go v0.28.4
+	// WatchList feature in client-go 0.35+ causes timeouts with fake clients.
+	// Keep at 0.34.4 until fakes support bookmark events for initial events stream.
```
csieber, line 42: Is there a bug we can reference?

csieber (approved):
thank you!

src/go/pkg/controller/approllout/controller.go
ensonic: Yes, previously the indentation was inconsistent :/

src/go.mod
ensonic: Reference bug.

I've found https://github.com/istio/istio/issues/59268 that has some workarounds, I can take a closer look in a followup.

PR #705, opened 2026-06-12
Add more context to error messages.

ensonic:
This will help debugging issues.

### DIFF
--- a/src/go/cmd/cr-syncer-auth-webhook/request.go
+++ b/src/go/cmd/cr-syncer-auth-webhook/request.go
@@ -8,8 +8,6 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-
-	"github.com/pkg/errors"
 )
 
 // Regex for RFC 1123 subdomain format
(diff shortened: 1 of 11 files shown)


src/go/cmd/cr-syncer-auth-webhook/request.go
```diff
@@ -49,7 +47,7 @@ func parseURL(urlString string) (*incomingRequest, error) {
 	//                             parts[0] parts[1] parts[2]   parts[3]   parts[4] parts[5]
 	parts := strings.Split(strings.TrimPrefix(url.Path, "/apis/core.kubernetes/apis/"), "/")
 	if len(parts) < 3 || len(parts) > 7 {
-		return nil, errors.New("unexpected URL length")
+		return nil, fmt.Errorf("unexpected URL path %q (split into %d parts)", url.Path, len(parts))
```
Tobias-Pe, line 50: Should we add how we split 

```suggestion
		return nil, fmt.Errorf("unexpected URL path %q (split into %d parts by '/')", url.Path, len(parts))
```

Tobias-Pe (approved):
Thanks!

src/go/cmd/cr-syncer-auth-webhook/request.go
ensonic: Too late :/. I'll send a quick folloup.

Tobias-Pe: Hehe no worries. Just a super small nit :)

PR #709, opened 2026-06-12
Increse test coverage for the relay

ensonic:
* Helper Functions: Added unit tests for marshalHeader, unmarshalHeader, and extractBackendNameAndPath, including edge cases and error conditions.
* Server Handlers:
  * Added tests for the health handler.
  * Improved error path coverage for serverRequestStream, serverResponse, and serverRequest (e.g., missing parameters, invalid proto bodies).
  * Added TestNewServerConfig to verify default and custom configurations.
  * Added TestResponseFilterChunking to cover asynchronous response streaming.
  * Added TestResponseTrailers and TestGrpcHeadersAsTrailers to verify trailer handling and conversion of "Grpc-" headers.
* Broker:
  * Added tests for RelayRequest error conditions (URL parsing errors, multiple clients for the same request ID).
  * Added TestGetRequestServerRestarting to cover server shutdown scenarios.
  * Ensured that inactivity timeouts and relay timeouts are handled correctly (though some tests are skipped in Short mode to avoid long delays).

### DIFF
--- a/src/go/cmd/http-relay-server/server/broker_test.go
+++ b/src/go/cmd/http-relay-server/server/broker_test.go
@@ -320,14 +320,14 @@ func TestReapWhileSendingRequest(t *testing.T) {
 	go func() {
 		req, reqErr = b.GetRequest(context.Background(), "foo", "/")
 		if reqErr != nil {
-			t.Errorf("GetRequest error:", reqErr)
+			t.Errorf("GetRequest error: %v", reqErr)
 		}
 		wg.Done()
 	}()
 	go func() {
 		_, respErr = b.RelayRequest("foo", &pb.HttpRequest{Id: proto.String(idOne), Url: proto.String("http://example.com/foo")})
 		if respErr != nil {
-			t.Errorf("RelayRequest error:", respErr)
+			t.Errorf("RelayRequest error: %v", respErr)
 		}
 		wg.Done()
 	}()
@@ -336,7 +336,7 @@ func TestReapWhileSendingRequest(t *testing.T) {
 		t.Errorf("Error making broker connection")
 	}
 
-	// start sending response to backend, BUT do not start consuming on the backend side
+	// start sending request stream to backend, BUT do not start consuming on the backend side
 	wg.Add(1)
 	go func() {
 		if ok := b.PutRequestStream(*req.Id, []byte(*req.Id)); !ok {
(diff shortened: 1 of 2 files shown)


src/go/cmd/http-relay-server/server/broker_test.go
```diff
@@ -351,3 +351,75 @@ func TestReapWhileSendingRequest(t *testing.T) {
...
+	}
+}
+
+func TestGetRequestServerRestarting(t *testing.T) {
+	b := newBroker()
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // Cancel immediately
+
+	_, err := b.GetRequest(ctx, "foo", "/")
+	if !errors.Is(err, ErrServerRestarting) {
```
csieber, line 389: `ErrServerRestarting` is not intuitive here. Can we add a comment why the cancelled context on the request results in this error?

src/go/cmd/http-relay-server/server/broker_test.go
```diff
@@ -351,3 +351,75 @@ func TestReapWhileSendingRequest(t *testing.T) {
...
+		t.Errorf("Expected ErrRelayTimeout, got %v", err)
+	}
+}
+
+func TestGetRequestStreamTimeout(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping slow timeout test")
+	}
+	b := newBroker()
+	b.req["foo"] = make(chan *pb.HttpRequest, 1)
```
csieber: Also add a comment there that the channel blocks indefinitely.

src/go/cmd/http-relay-server/server/server_test.go
```diff
@@ -186,6 +188,101 @@ func TestClientHandlerWithChunkedResponse(t *testing.T) {
...
+
+	marshaled := marshalHeader(&h)
+	if len(marshaled) != 3 {
+		t.Errorf("Expected 3 marshaled headers, got %d", len(marshaled))
+	}
+
+	w := httptest.NewRecorder()
+	unmarshalHeader(w, marshaled)
+
+	if got := w.Header().Values("X-Test"); len(got) != 2 || got[0] != "value1" || got[1] != "value2" {
```
csieber, line 206: Does this assume the header order is stable? Is this true?

src/go/cmd/http-relay-server/server/server_test.go
```diff
@@ -427,3 +524,185 @@ func BenchmarkCreateId(b *testing.B) {
...
+	})
+
+	wg.Wait()
+	resp := respRecorder.Result()
+	if got := resp.Trailer.Get("X-Trailer"); got != "value" {
+		t.Errorf("Expected trailer X-Trailer: value, got %q", got)
+	}
+}
+
+func TestGrpcHeadersAsTrailers(t *testing.T) {
```
csieber: This test case is very similar to the previous one. Do we really need both?

csieber (approved):

src/go/cmd/http-relay-server/server/broker_test.go
ensonic: I think I should rename the error. There is nothing restarting.
The error is returned to serverRequest()
https://github.com/googlecloudrobotics/core/blob/main/src/go/cmd/http-relay-server/server/server.go#L481

Okay if I do the rename in a followup?

src/go/cmd/http-relay-server/server/broker_test.go
ensonic: Done. We shoudl consider to also add a context where.

src/go/cmd/http-relay-server/server/broker_test.go
csieber: SGTM

src/go/cmd/http-relay-server/server/server_test.go
ensonic: I think it is (it is a slice). Also this tests the marshalling that preserves the order.

src/go/cmd/http-relay-server/server/server_test.go
ensonic: merged both into a table driven test. Better?

csieber: looks good!

PR #710, opened 2026-06-12
Add syntax highlight hints to README.md

ensonic:
Add language tags where missing. Also use "shell" instead of "bash".

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -12,21 +12,21 @@ BUILD files for Go.
 
 To automatically update dependencies in bazel BUILD files run:
 
-```
+```shell
 bazel run //:gazelle
 ```
 
 To re-generate Go modules dependencies run this from the top-level source
 directory:
 
-```
+```shell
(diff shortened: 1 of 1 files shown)


src/README.md
```diff
@@ -42,13 +42,13 @@ More tips on this [one-pager](https://encore.dev/guide/go.mod)
...
 
-```
+```shell
 go install github.com/google/go-licenses@latest
 ```
 
 run it:
 
-```
+```shell
```
Tobias-Pe, line 51: Shell vs bash in markdown. I usually see more bash but feel free to ignore this comment if shell works

Tobias-Pe (approved):

src/README.md
ensonic:
```shell
git grep "\`\`\`shell" | wc -l
114
git grep "\`\`\`bash" | wc -l
15
```
I'll change the other as a followup.

PR #711, opened 2026-06-12
Consistently use "shell" in markdown files

ensonic:
There is no visual difference, but hey - lets be consitent.


Tobias-Pe (approved):
Thanks!

PR #712, opened 2026-06-15
Extend echo server to accept path based and query parameters 

liyin00:
Add dynamic status echoing, graceful shutdown, and K8s scaling configs

This PR updates the status-echo-server to support dynamic status code
extraction from the request path or query parameters, and adds
reliability and scalability configurations for deployment.

Key Changes:

1. Dynamic Status Code Extraction:
   - Extracts status codes from the path suffix (e.g., `/status/201`)
     or the `status` query parameter (e.g., `?status=202`).
   - Path-based status codes take precedence (e.g.,
     `/status/200?status=500` returns 200).
   - Implemented using efficient string manipulation instead of regex
     to minimize memory allocations under heavy load.

2. Graceful Shutdown & Health Probes:
   - Implemented a graceful shutdown sequence with a 5-second readiness
     drain delay to prevent dropping active connections during Pod
     termination.
   - Added a `/healthz` endpoint for liveness and readiness probes
     (returns 503 Service Unavailable during the shutdown phase).
   - Added integration tests verifying shutdown behavior and probe
     states.

3. Kubernetes Scaling & Reliability (istio.yaml):
   - Set optimal resource specs (1.0 CPU Request/Limit, 32Mi Request /
     64Mi Limit) verified by benchmarks to support 10k Requests Per
     Second (RPS) per pod.
   - Added Horizontal Pod Autoscaler (HPA) targeting 80% CPU
     utilization (min 2, max 10 replicas).
   - Added Pod Disruption Budget (PDB) allowing a maximum of 1
     unavailable pod for maintenance reliability.

Benchmarking results as attached.

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -129,7 +129,7 @@ spec:
         - --port=8080
         - --http_status=404
         readinessProbe:
-          httpGet: 
+          httpGet:
             path: /readyz
             port: 8080
             scheme: HTTP
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -110,7 +110,7 @@ metadata:
   labels:
     app: crc-auth-gateway-catchall-svc
 spec:
-  replicas: 1
+  replicas: 2
   selector:
     matchLabels:
       app: crc-auth-gateway-catchall-svc
@@ -130,11 +130,19 @@ spec:
         - --port=8080
         - --http_status=404
         readinessProbe:
-          tcpSocket:
+          httpGet: 
+            path: /healthz
             port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 2 
+          periodSeconds: 5
         livenessProbe:
-          tcpSocket:
+          httpGet: 
+            path: /healthz
             port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 2 
+          periodSeconds: 5
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
@@ -148,6 +156,52 @@ spec:
         runAsNonRoot: true
         runAsUser: 65532
         runAsGroup: 65532
+      resources:
+        requests:
+          cpu: "1.0"
+          memory: "32Mi"
+        limits:
+          cpu: "1.5"
+          memory: "64Mi"
+---
+{{- if eq .Values.use_istio "true" }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: crc-auth-gateway-catchall-pdb
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      app: crc-auth-gateway-catchall-svc
+{{- end}}
+---
+{{- if eq .Values.use_istio "true" }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: crc-auth-gateway-catchall-hpa
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: crc-auth-gateway-catchall-svc
+  minReplicas: 2
+  maxReplicas: 10
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: 80
+  - type: Resource
+    resource:
+      name: memory
+      target:
+        type: Utilization
+        averageUtilization: 80
+{{- end}}
 ---
 apiVersion: v1
 kind: Service
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -110,7 +110,6 @@ metadata:
   labels:
     app: crc-auth-gateway-catchall-svc
 spec:
-  replicas: 2
   selector:
     matchLabels:
       app: crc-auth-gateway-catchall-svc
@@ -143,6 +142,13 @@ spec:
             scheme: HTTP
           initialDelaySeconds: 2 
           periodSeconds: 5
+        resources:
+          requests:
+            cpu: "1.0"
+            memory: "32Mi"
+          limits:
+            cpu: "1.5"
+            memory: "64Mi"
         securityContext:
           allowPrivilegeEscalation: false
           readOnlyRootFilesystem: true
@@ -156,13 +162,7 @@ spec:
         runAsNonRoot: true
         runAsUser: 65532
         runAsGroup: 65532
-      resources:
-        requests:
-          cpu: "1.0"
-          memory: "32Mi"
-        limits:
-          cpu: "1.5"
-          memory: "64Mi"
+
 ---
 apiVersion: policy/v1
 kind: PodDisruptionBudget
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -164,7 +164,6 @@ spec:
           cpu: "1.5"
           memory: "64Mi"
 ---
-{{- if eq .Values.use_istio "true" }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
@@ -174,9 +173,7 @@ spec:
   selector:
     matchLabels:
       app: crc-auth-gateway-catchall-svc
-{{- end}}
 ---
-{{- if eq .Values.use_istio "true" }}
 apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
@@ -201,7 +198,6 @@ spec:
       target:
         type: Utilization
         averageUtilization: 80
-{{- end}}
 ---
 apiVersion: v1
 kind: Service
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -169,7 +169,7 @@ kind: PodDisruptionBudget
 metadata:
   name: crc-auth-gateway-catchall-pdb
 spec:
-  minAvailable: 1
+  maxUnavailable: 33%
   selector:
     matchLabels:
       app: crc-auth-gateway-catchall-svc
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -130,7 +130,7 @@ spec:
         - --http_status=404
         readinessProbe:
           httpGet: 
-            path: /healthz
+            path: /readyz
             port: 8080
             scheme: HTTP
           initialDelaySeconds: 2 
@@ -169,7 +169,7 @@ kind: PodDisruptionBudget
 metadata:
   name: crc-auth-gateway-catchall-pdb
 spec:
-  maxUnavailable: 1
+  minAvailable: 1
   selector:
     matchLabels:
       app: crc-auth-gateway-catchall-svc
@@ -189,12 +189,6 @@ spec:
   - type: Resource
     resource:
       name: cpu
-      target:
-        type: Utilization
-        averageUtilization: 80
-  - type: Resource
-    resource:
-      name: memory
       target:
         type: Utilization
         averageUtilization: 80
--- a/src/go/cmd/status-echo-server/main.go
+++ b/src/go/cmd/status-echo-server/main.go
@@ -99,7 +99,7 @@ func determineStatusCode(req *http.Request, defaultStatus int) int {
 }
 
 func healthzHandler(w http.ResponseWriter, req *http.Request) {
-	fmt.Fprintln(w, "OK")
+	io.WriteString(w, "OK\n")
 }
 
 func readyzHandler(w http.ResponseWriter, req *http.Request) {
@@ -107,7 +107,7 @@ func readyzHandler(w http.ResponseWriter, req *http.Request) {
 		http.Error(w, "Server shutting down", http.StatusServiceUnavailable)
 		return
 	}
-	fmt.Fprintln(w, "OK")
+	io.WriteString(w, "OK\n")
 }
 
 func main() {
--- a/src/go/cmd/status-echo-server/main.go
+++ b/src/go/cmd/status-echo-server/main.go
@@ -37,6 +37,7 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	"io"
 
 	"github.com/googlecloudrobotics/ilog"
 )
@@ -59,7 +60,7 @@ func echoHandler(w http.ResponseWriter, req *http.Request) {
 
 	statusCode := determineStatusCode(req, *httpStatus)
 	w.WriteHeader(statusCode)
-	w.Write([]byte(http.StatusText(statusCode)))
+	io.WriteString(w, http.StatusText(statusCode))
 }
 
 func determineStatusCode(req *http.Request, defaultStatus int) int {
@@ -76,7 +77,7 @@ func determineStatusCode(req *http.Request, defaultStatus int) int {
 			statusStr = suffix
 		}
 	}
-	if statusStr == "" {
+	if (statusStr == "") && (req.URL.RawQuery != "") {
 		statusStr = req.URL.Query().Get("status")
 	}
 
@@ -97,8 +98,11 @@ func determineStatusCode(req *http.Request, defaultStatus int) int {
 	return statusCode
 }
 
-// Using a single endpoint for both health and readiness probes since there are no external dependencies for now. 
 func healthzHandler(w http.ResponseWriter, req *http.Request) {
+	fmt.Fprintln(w, "OK")
+}
+
+func readyzHandler(w http.ResponseWriter, req *http.Request) {
 	if isShuttingDown.Load() {
 		http.Error(w, "Server shutting down", http.StatusServiceUnavailable)
 		return
@@ -117,6 +121,7 @@ func main() {
 
 	mux := http.NewServeMux()
 	mux.HandleFunc("/healthz", healthzHandler)
+	mux.HandleFunc("/readyz", readyzHandler)
 	mux.HandleFunc("/", echoHandler)
 
 	addr := fmt.Sprintf(":%d", *port)
@@ -153,7 +158,7 @@ func shutdownServer(srv *http.Server, forceCancel context.CancelFunc) {
 	time.Sleep(readinessDrainDelay)
 
 	// Attempt a graceful shutdown (finish processing of current requests).
-	slog.Info("Initiating server shutdown with timeout of", shutdownPeriod)
+	slog.Info("Initiating server shutdown", "timeout", shutdownPeriod)
 	shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownPeriod)
 	defer cancel()
 
--- a/src/go/cmd/status-echo-server/main.go
+++ b/src/go/cmd/status-echo-server/main.go
@@ -85,10 +85,10 @@ func determineStatusCode(req *http.Request, defaultStatus int) int {
 	if statusStr != "" {
 		parsedCode, err := strconv.Atoi(statusStr)
 		if err != nil {
-			slog.Warn("status code could not be converted to integer", "error", err)
+			slog.Warn("Status code could not be converted to integer", "error", err)
 			statusCode = http.StatusBadRequest
 		} else if http.StatusText(parsedCode) == "" {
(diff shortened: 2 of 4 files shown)


liyin00:
# Benchmark Results: status-echo-server 10k RPS

This document summarizes the benchmark results to find the optimal CPU and memory requests/limits for a single replica of `status-echo-server` to support 10,000 Requests Per Second (RPS).

## Executive Summary

Based on the benchmarks, the optimal resource configuration for a single replica to support 10k RPS is:

*   **Optimal CPU Limit:** **1.0 Core** (Absolute minimum: ~0.9 Cores)
*   **Optimal Memory Limit:** **16MB** (Absolute minimum tested: 12MB, but 16MB provides a safer margin)

With this configuration, the server achieved **9,999+ QPS** with **p99 latency < 1.0ms** and **0% error rate**.

---

## Methodology

*   **Server Build:** Built using Bazel (`//src/go/cmd/status-echo-server:status-echo-server-image`).
*   **Deployment:** Run locally in Docker with resource limits (`--cpus` and `--memory`).
*   **Load Generator:** `fortio/fortio` running in a separate container using host network.
*   **Test Case:** HTTP GET requests to `/status/200` (returns static "OK" response).
*   **Load Profile:** 10,000 target RPS for 15 seconds (after 2 seconds warmup).
*   **Monitoring:** `docker stats` polled every 0.5s to capture maximum CPU and Memory usage.

---

## Detailed Results

### Phase 1: CPU Limit Search (Memory fixed at 128MB)

We tested various CPU limits to find the point where performance degrades.

| CPU Limit (Cores) | Target RPS | Actual QPS | p50 Latency (ms) | p99 Latency (ms) | Max CPU Usage | Max Mem (MB) | Error Rate | Status |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 2.0 | 10,000 | 9,998.7 | 0.54 | 0.99 | 92.8% (of 1 core) | 14.7 | 0.00% | OK |
| 1.5 | 10,000 | 9,998.9 | 0.55 | 0.99 | 92.6% (of 1 core) | 16.0 | 0.00% | OK |
| 1.2 | 10,000 | 9,999.0 | 0.54 | 0.99 | 93.2% (of 1 core) | 13.8 | 0.00% | OK |
| **1.0** | **10,000** | **9,999.3** | **0.55** | **0.99** | **89.1% (of 1 core)** | **12.6** | **0.00%** | **Optimal** |
| 0.8 | 10,000 | 9,427.2 | 0.54 | 1.00 | 81.1% (of 1 core) | 14.2 | 0.00% | Throttled |
| 0.5 | 10,000 | 5,812.6 | 0.55 | 1.00 | 50.9% (of 1 core) | 13.3 | 0.00% | Severely Throttled |

**Conclusion (CPU):** A CPU limit of 1.0 core is required to fully support 10k RPS. At 0.8 cores, the server is throttled and can only deliver ~9.4k QPS.

---

### Phase 2: Memory Limit Search (CPU fixed at 1.0 Core)

With CPU fixed at 1.0 core, we reduced memory limits to find the crashing threshold (OOM).

| Memory Limit | Target RPS | Actual QPS | p50 Latency (ms) | p99 Latency (ms) | Max CPU Usage | Max Mem (MB) | Error Rate | Status |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 128MB | 10,000 | 9,999.7 | 0.54 | 0.99 | 93.3% | 13.7 | 0.00% | OK |
| 64MB | 10,000 | 9,999.3 | 0.55 | 0.99 | 90.3% | 14.5 | 0.00% | OK |
| 32MB | 10,000 | 9,998.5 | 0.54 | 0.99 | 92.8% | 13.4 | 0.00% | OK |
| 24MB | 10,000 | 9,999.0 | 0.54 | 0.99 | 89.2% | 12.8 | 0.00% | OK |
| **16MB** | **10,000** | **9,999.0** | **0.55** | **0.99** | **92.2%** | **13.8** | **0.00%** | **Recommended** |
| 12MB | 10,000 | 9,999.0 | 0.55 | 0.99 | 91.4% | 11.9 | 0.00% | OK (Tight) |

**Conclusion (Memory):** The server has a very small memory footprint (~12-15MB under load). It ran successfully even with a 12MB limit, but 16MB (or 32MB) is recommended for production to avoid potential OOMs during garbage collection spikes or minor traffic variations.

---

## Production Recommendations

For deployment in Kubernetes, we recommend the following settings for a single replica targeting 10k RPS:

```yaml
resources:
  requests:
    cpu: "1.0"
    memory: "16Mi"
  limits:
    cpu: "1.0"      # Prevent throttling if it goes slightly over
    memory: "32Mi"  # Allow some buffer for GC spikes 
```
Note: I doubled the memory just to have more headroom. So the memory requests is 32Mi and limits is 64Mi on the manifests. 

*   **CPU Request:** 1.0 is critical to ensure the replica gets enough CPU cycles to handle the load.
*   **Memory Request/Limit:** 16Mi/32Mi is extremely safe for this server, given its actual usage is ~14MB.

---

## Alternative: Estimating Resources with Go `pprof`

We can also estimate these requirements by profiling an unconstrained server under 10k RPS load using Go's native `pprof` tool.

### CPU Estimation

By running the server without limits and capturing a CPU profile during the 10k RPS load test:

```bash
curl -o cpu.pprof http://localhost:8080/debug/pprof/profile?seconds=10
go tool pprof -text cpu.pprof
```

The profile header shows:
*   **Duration:** 10.14s
*   **Total samples:** 9.91s (97.72%)

**Analysis:**
The profile shows that the server consumed **9.91 seconds of CPU time** over a **10.14 second real-time window**.

$$\text{Estimated Cores} = \frac{\text{Total CPU Time}}{\text{Real-time Duration}} = \frac{9.91\text{s}}{10.14\text{s}} \approx 0.98\text{ Cores}$$

This confirms our container test result: the server requires almost exactly **1.0 core** of CPU capacity to handle 10k RPS. Any limit below 1.0 will cause throttling.

### Memory Estimation

Capturing a heap profile under load:

```bash
curl -o heap.pprof http://localhost:8080/debug/pprof/heap
go tool pprof -text heap.pprof
```

Shows:
*   **In-use Heap:** ~4.7 MB (includes ~1.7 MB of pprof overhead)
*   **Net In-use Heap:** ~3.0 MB

**Analysis:**
While `pprof` shows the live heap size is very small (~3MB), it does not account for Go runtime overhead (stacks, allocator cache, descriptor tables, etc.) which is visible in RSS (resident set size).
Our container benchmarks showed actual memory usage was ~14MB.
Therefore, while `pprof` heap profile gives a baseline of **3-5MB**, we must add a safety margin of at least 10MB to account for runtime overhead and GC workspace. This aligns with our recommended **16MB - 32MB** limit.

liyin00:
Just a heads-up: I set it up to ignore status codes if they're in the middle of the path. So, something like http://localhost:8080/status/200/super/simple/service will just fall back to the default. Let me know if I should change this!

src/app_charts/base/cloud/istio.yaml
```diff
@@ -148,6 +156,52 @@ spec:
...
         runAsGroup: 65532
+      resources:
+        requests:
+          cpu: "1.0"
+          memory: "32Mi"
+        limits:
+          cpu: "1.5"
+          memory: "64Mi"
+---
+{{- if eq .Values.use_istio "true" }}
```
ensonic: the whole file is wrapped into `{{- if eq .Values.use_istio "true" }}`

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,153 @@
...
+	"syscall"
 	"time"
 
 	"github.com/googlecloudrobotics/ilog"
 )
 
+const (
+	_shutdownPeriod      = 15 * time.Second
+	_shutdownHardPeriod  = 3 * time.Second
+	_readinessDrainDelay = 5 * time.Second
```
ensonic: Can we not use '_', as symbols starting with a lowercase char are already package private

ensonic (approved):
Thanks!

src/go/cmd/status-echo-server/main.go
liyin00: Updated, thanks!

src/app_charts/base/cloud/istio.yaml
liyin00: Removed, thanks!

src/app_charts/base/cloud/istio.yaml
```diff
@@ -110,7 +110,7 @@ metadata:
   labels:
     app: crc-auth-gateway-catchall-svc
 spec:
-  replicas: 1
+  replicas: 2
```
Tobias-Pe: [ARCHITECTURE] Prefer omitting the `replicas` field from the Deployment spec when using a `HorizontalPodAutoscaler`. This prevents conflicting replica counts from being applied and overriding active scaling decisions during deployment updates.

🤖 **AI Reviewer** (Jetski) | Traces: architecture: go/traj/71856814-6b3b-4e41-a765-089de8dc9c62

src/app_charts/base/cloud/istio.yaml
```diff
@@ -148,6 +156,48 @@ spec:
         runAsNonRoot: true
         runAsUser: 65532
         runAsGroup: 65532
+      resources:
```
Tobias-Pe: [CORRECTNESS] The `resources` field is defined under the Pod spec (at the same level as `securityContext`), which is invalid. It must be defined under the container spec (e.g., under `containers[0]` with 8 spaces of indentation).

🤖 **AI Reviewer** (Jetski) | Traces: correctness: go/traj/0a25bbfc-07f3-437c-825e-f7f0212caaa9

src/go/cmd/status-echo-server/main.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package main runs a multiplexing HTTP relay server.
-//
+// Package main runs a simple HTTP server that echoes back a configured
```
Tobias-Pe, line 15: [CORRECTNESS] [Reviewer Note: Comment originally intended for line 190 (outside diff hunk)] The `net/http/pprof` package is imported for side-effects, but it registers handlers on `http.DefaultServeMux`. Since the server uses a custom `http.NewServeMux()`, the pprof endpoints will not be accessible. Prefer registering the pprof handlers manually on the custom mux or removing the import.

🤖 **AI Reviewer** (Jetski) | Traces: correctness: go/traj/0a25bbfc-07f3-437c-825e-f7f0212caaa9

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,152 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
+	_ "net/http/pprof"
```
Tobias-Pe: [ARCHITECTURE, STYLE] The `_ "net/http/pprof"` import registers handlers on the default `ServeMux`, but this server uses a custom `ServeMux`, making these handlers unreachable.

Please consider one of the following options:
* **Remove the import** if profiling is not needed.
* **Register the pprof handlers** on the custom `ServeMux` explicitly.
* **Mount `http.DefaultServeMux`** on the custom mux to expose the default handlers.

🤖 **AI Reviewer** (Jetski) | Traces: style: go/traj/4c394c78-eedd-467f-b108-cec64cff7be1, architecture: go/traj/71856814-6b3b-4e41-a765-089de8dc9c62

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,152 @@
...
 
 func echoHandler(w http.ResponseWriter, req *http.Request) {
 	slog.Debug("echo", slog.String("path", req.URL.Path))
 	w.Header().Set("Content-Type", "text/plain")
-	w.WriteHeader(*httpStatus)
-	w.Write([]byte(http.StatusText(*httpStatus)))
+
+	statusCode := determineStatusCode(req, *httpStatus)
+	w.WriteHeader(statusCode)
+	w.Write([]byte(http.StatusText(statusCode)))
```
Tobias-Pe: [PERFORMANCE] Prefer using `io.WriteString(w, http.StatusText(statusCode))` instead of `w.Write([]byte(...))`. This avoids allocating a new byte slice for the status text, as standard `http.ResponseWriter` implementations support writing strings directly.

🤖 **AI Reviewer** (Jetski) | Traces: performance: go/traj/b641ff3f-85a8-4518-8867-4c4e7754666e

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,152 @@
...
+	// Try to extract status code from path suffix (e.g., /status/200) first, falling back to the "status" query parameter if not found.
+	if idx := strings.LastIndex(path, "/status/"); idx != -1 {
+		suffix := path[idx+len("/status/"):]
+		// Ensure there are no more slashes after "/status/" to guarantee the pattern is strictly .../status/<something>$
+		if !strings.Contains(suffix, "/") {
+			statusStr = suffix
+		}
+	}
+	if statusStr == "" {
+		statusStr = req.URL.Query().Get("status")
```
Tobias-Pe, line 81: [PERFORMANCE] Prefer checking if `req.URL.RawQuery` is empty before calling `req.URL.Query().Get(...)`. `Query()` parses the raw query and allocates a new map on every call, which can be avoided if no query parameters are present.

🤖 **AI Reviewer** (Jetski) | Traces: performance: go/traj/b641ff3f-85a8-4518-8867-4c4e7754666e

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,152 @@
...
+		} else {
+			statusCode = parsedCode
+		}
+	}
+
+	return statusCode
+}
+
+// Using a single endpoint for both health and readiness probes since there are no external dependencies for now. 
+func healthzHandler(w http.ResponseWriter, req *http.Request) {
```
Tobias-Pe, line 101: [ARCHITECTURE] Prefer separating the liveness and readiness endpoints (e.g., using `/healthz` and `/readyz`) instead of using a single shared `/healthz` endpoint. Returning 503 during shutdown will fail the liveness probe, which can trigger premature pod termination before graceful shutdown completes.

🤖 **AI Reviewer** (Jetski) | Traces: architecture: go/traj/71856814-6b3b-4e41-a765-089de8dc9c62

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,152 @@
...
 }
+
+func shutdownServer(srv *http.Server, forceCancel context.CancelFunc) {
+	// Fail health checks to stop routing of new traffic to terminating pod. 
+	isShuttingDown.Store(true)
+	slog.Info("Received shutdown signal. Failing health check and waiting for drain delay.")
+	time.Sleep(readinessDrainDelay)
+
+	// Attempt a graceful shutdown (finish processing of current requests).
+	slog.Info("Initiating server shutdown with timeout of", shutdownPeriod)
```
Tobias-Pe: [STYLE] Prefer using key-value pairs or structured attributes for `slog.Info` (e.g., `slog.Info("Initiating server shutdown", "timeout", shutdownPeriod)`) instead of passing the duration as a loose argument, which results in a '!BADKEY' key in the structured log output.

🤖 **AI Reviewer** (Jetski) | Traces: style: go/traj/4c394c78-eedd-467f-b108-cec64cff7be1

Tobias-Pe (changes requested):

Tobias-Pe (commented):
Changing to comment only to not bock. Thanks for the thorough analysis and benchmarking, look really good!

Tobias-Pe (commented):
EDIT: hm changing from request 4 change to comment only not possible. Sorry!

src/app_charts/base/cloud/istio.yaml
```diff
@@ -148,6 +162,42 @@ spec:
...
+  minReplicas: 2
+  maxReplicas: 10
+  metrics:
+  - type: Resource
+    resource:
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: 80
+  - type: Resource
```
awhdesmond: I believe scaling based on memory might not be required.. the memory requests/limits are quite low.

src/app_charts/base/cloud/istio.yaml
```diff
@@ -148,6 +162,42 @@ spec:
...
         runAsUser: 65532
         runAsGroup: 65532
+
+---
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: crc-auth-gateway-catchall-pdb
+spec:
+  maxUnavailable: 1
```
awhdesmond: Maybe we can use percentages such as `maxUnavailable: 33%` instead, so we can scale proportionally with HPA?

src/app_charts/base/cloud/istio.yaml
liyin00: Removed, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: Removed as we don't need to profile the server live, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: Removed, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: Ahh true, separated it out. Good catch!

src/app_charts/base/cloud/istio.yaml
liyin00: Removed, thanks!

src/app_charts/base/cloud/istio.yaml
liyin00: Updated, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: amended, thanks

src/go/cmd/status-echo-server/main.go
liyin00: amended, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: removed altogether as we are not profiling on live servers. thanks for this!

src/app_charts/base/cloud/istio.yaml
liyin00: indented and rebuilt to verify. thanks for catching this!

awhdesmond (approved):

src/go/cmd/status-echo-server/main.go
```diff
@@ -24,53 +24,156 @@
...
+		} else {
+			statusCode = parsedCode
+		}
+	}
+
+	return statusCode
+}
+
+func healthzHandler(w http.ResponseWriter, req *http.Request) {
+	fmt.Fprintln(w, "OK")
```
Tobias-Pe: [PERFORMANCE] Consider using `io.WriteString(w, "OK\n")` instead of `fmt.Fprintln` for the health check endpoint. This avoids interface allocation and formatting overhead.

🤖 **AI Reviewer** (Jetski) | Traces: performance: go/traj/e6984bf0-4beb-4d25-b72d-029fc08fd730

src/app_charts/base/cloud/istio.yaml
```diff
@@ -130,11 +129,26 @@ spec:
         - --port=8080
         - --http_status=404
         readinessProbe:
-          tcpSocket:
+          httpGet:
```
Tobias-Pe: [STYLE] Please remove trailing whitespace to maintain clean formatting.

🤖 **AI Reviewer** (Jetski) | Traces: style: go/traj/c4021ccb-eca2-4969-9b21-b8290d0b21d0

Tobias-Pe: nit

src/go/cmd/status-echo-server/main.go
Tobias-Pe: Not sure if this is true, so classify as nit

Tobias-Pe (approved):
Thanks!

LGTM GIF

src/app_charts/base/cloud/istio.yaml
liyin00: updated, thanks!

src/go/cmd/status-echo-server/main.go
liyin00: updated, thanks!

PR #715, opened 2026-06-15
Code style cleanups

ensonic:
* use one-line error handling form with limited scope where possible.
* return structs as we initialize them (without intermediate vars)
* use swicth case over multiple if else blocks.

### DIFF
--- a/src/go/cmd/app-rollout-controller/main.go
+++ b/src/go/cmd/app-rollout-controller/main.go
@@ -68,7 +68,7 @@ func main() {
 	}
 
 	if err := runController(ctx, kubernetesConfig, helmParams); err != nil {
-		slog.Error("Exit", ilog.Err(runController(ctx, kubernetesConfig, helmParams)))
+		slog.Error("Exit", ilog.Err(err))
 		os.Exit(1)
 	}
 	slog.Info("Exit")
--- a/src/go/cmd/http-relay-server/server/server.go
+++ b/src/go/cmd/http-relay-server/server/server.go
@@ -268,8 +268,7 @@ func (s *Server) bidirectionalStream(backendCtx backendContext, w http.ResponseW
 			// This must be a new buffer each time, as the channel is not making a copy
 			bytes := make([]byte, s.conf.BlockSize)
 			// Here we get the client stream (e.g. kubectl or k9s)
-			n, err := bufrw.Read(bytes)
-			if err != nil {
+			if n, err := bufrw.Read(bytes); err != nil {
 				if errors.Is(err, net.ErrClosed) {
 					// Request ended and connection closed by HTTP server.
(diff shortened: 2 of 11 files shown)


src/go/cmd/app-rollout-controller/main.go
```diff
@@ -68,7 +68,7 @@ func main() {
 	}
 
 	if err := runController(ctx, kubernetesConfig, helmParams); err != nil {
-		slog.Error("Exit", ilog.Err(runController(ctx, kubernetesConfig, helmParams)))
+		slog.Error("Exit", ilog.Err(err))
```
csieber, line 71: Are we sure this has the same behavior as before? Before it was calling runController a second time.

src/go/cmd/http-relay-server/server/server.go
```diff
@@ -278,19 +277,20 @@ func (s *Server) bidirectionalStream(backendCtx backendContext, w http.ResponseW
 					slog.Error("Error reading from bidi-stream", slog.String("ID", backendCtx.Id), ilog.Err(err))
 				}
 				return
+			} else {
```
csieber, line 280: Do we need the else here? We have the return above

src/go/cmd/metadata-server/metadata.go
```diff
@@ -141,7 +140,8 @@ func fromAcceptedIP(w http.ResponseWriter, r *http.Request, allowedSources *net.
 		http.Error(w, "Unable to check authorization", http.StatusInternalServerError)
 		return false
 	}
-	if ip := net.ParseIP(ipPort[0]); ip == nil || !allowedSources.Contains(ip) {
+	ip := net.ParseIP(ipPort[0])
+	if ip == nil || !allowedSources.Contains(ip) {
```
csieber: nit: Are you worried the line is too long? Before one line was a bit more concise

csieber (approved):
thanks!

src/go/cmd/metadata-server/metadata.go
ensonic: That was a mistake that slipped. reverted

src/go/cmd/http-relay-server/server/server.go
ensonic: We need it if we want to access "n" as a local var. But if that makes the code less readable, I can also revert this change.

src/go/cmd/app-rollout-controller/main.go
ensonic: I think it was a copy and paste mistake: see
https://github.com/googlecloudrobotics/core/commit/95cd2d2ecf566a4f7007fda4e1f91e1966ebc751

ensonic: @csieber if you feel better, I can extract this and send it separately to @therjak

csieber (approved):

PR #728, opened 2026-06-16
Deploy nginx ingress shield to mitigate HTTP/2 HPACK exploit

awhdesmond:
This PR creates the resources for provisioning an NGINX proxy and daisy-chain traffic from it to the internal legacy ingress-nginx. 

- Deploy nginx ingress shield to mitigate HTTP/2 HPACK exploit. 
- Gated behind `.Values.use_nginx_shield`

Will not affect existing traffic as we have not yet switch the existing `ingress-nginx` k8s service selector. That will come later using kubectl patching in `deploy.sh`

### DIFF
--- a/src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
+++ b/src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
@@ -16,30 +16,6 @@ spec:
         k8s-app: nginx-ingress-controller-shield
     spec:
       serviceAccountName: ingress-nginx
-      shareProcessNamespace: true
-      initContainers:
-      - name: cert-loader
-        image: alpine/k8s:1.29.2
-        restartPolicy: Always
-        command: ["/bin/sh", "-c", "/scripts/loader.sh"]
-        volumeMounts:
-        - name: certs-nginx
-          mountPath: /etc/nginx/certs
-        - name: script
-          mountPath: /scripts
-        startupProbe:
-          exec:
-            command: ["test", "-f", "/etc/nginx/certs/wildcard-mappings.map"]
-          initialDelaySeconds: 1
-          periodSeconds: 2
-          failureThreshold: 30
-        resources:
-          requests:
-            cpu: 50m
-            memory: 64Mi
-          limits:
-            cpu: 100m
-            memory: 128Mi
       containers:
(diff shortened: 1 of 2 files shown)


src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
```diff
@@ -0,0 +1,544 @@
...
+      # Create a temporary file for the new map
+      NEW_MAP_FILE=$(mktemp)
+      
+      # Track synced filenames in this run to detect deletions
+      synced_files=""
+      changed=false
+      
+      # Find all TLS secrets across all namespaces and retrieve their data in one list operation
+      # Tab-separated values: namespace <tab> name <tab> base64_crt <tab> base64_key
+      secrets_data=$(kubectl get secrets --all-namespaces --field-selector type=kubernetes.io/tls -o jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.metadata.name}{"\t"}{.data.tls\.crt}{"\t"}{.data.tls\.key}{"\n"}{end}')
```
ensonic:
```
kubectl get secrets --all-namespaces --field-selector type=kubernetes.io/tls
NAMESPACE   NAME                              TYPE                DATA   AGE
default     app-rollout-controller-tls        kubernetes.io/tls   3      221d
default     chart-assignment-controller-tls   kubernetes.io/tls   3      221d
default     cluster-authority                 kubernetes.io/tls   3      221d
default     example-tls                       kubernetes.io/tls   2      31d
default     tls                               kubernetes.io/tls   2      221d
```
I think we only need the `NAME=tls` one. The other ones are mostly for the api server to provide apis for the CRDs.
Sadly there is no unique label to differentiate them (e.g. based on parent cert).

ensonic (commented):
Thanks

ensonic (approved):

src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
awhdesmond: Thanks! I removed the logic to handle cross-namespace certificates and updated the script to reload NGINX whenever the default tls cert is renewed.

ensonic (approved):

PR #731, opened 2026-06-16
Regenerate k8s client libraries.

ensonic:
Go from 1.22 to 1.28. I've bumped into some errors with 1.30 and for >=1.32 we need to update the script.


src/go/pkg/client/informers/factory.go
```diff
@@ -106,20 +111,39 @@ func NewSharedInformerFactoryWithOptions(client versioned.Interface, defaultResy
...
 	defer f.lock.Unlock()
 
+	if f.shuttingDown {
+		return
+	}
+
 	for informerType, informer := range f.informers {
 		if !f.startedInformers[informerType] {
-			go informer.Run(stopCh)
+			f.wg.Add(1)
```
csieber, line 124: Somehow these changes do not match the PR description. Are you sure they should be part of this change? Also `informer := informer` should not be necessary anymore since 1.22 (https://go.dev/blog/loopvar-preview).
If we use waitgroups, I think wg.Go() is nicer syntax.

ensonic: These files are generated by the generate.sh script. I would rather not patch the generated files.
I've deployed it to my test project and e.g. the chart assignments work.

csieber (approved):

PR #734, opened 2026-06-17
Deploy official Istio dashboards in update-istio.sh

awhdesmond:
Deploy [official Istio dashboards](https://github.com/istio/istio/tree/1.29.4/manifests/addons/dashboards)based on the version used in `update-istio.sh`. I think we can start with these and add custom dashboards when required.

Grafana version required: "pluginVersion": "v11.0.0"

### DIFF
--- a/third_party/istio/update-istio.sh
+++ b/third_party/istio/update-istio.sh
@@ -65,10 +65,21 @@ DASHBOARDS=(
 dashboard_yaml="${tmpdir}/dashboards.yaml"
 true > "${dashboard_yaml}"
 
+curl_args=()
+for dashboard in "${DASHBOARDS[@]}"; do
+  curl_args+=(
+    -o "${tmpdir}/${dashboard}"
+    "https://raw.githubusercontent.com/istio/istio/${VERSION}/manifests/addons/dashboards/${dashboard}"
+  )
+done
+
+if ! curl -fsSL "${curl_args[@]}"; then
+  echo "Failed to download dashboards" >&2
+  exit 1
+fi
+
 for dashboard in "${DASHBOARDS[@]}"; do
   json_file="${tmpdir}/${dashboard}"
-  curl -fsSL "https://raw.githubusercontent.com/istio/istio/${VERSION}/manifests/addons/dashboards/${dashboard}" -o "${json_file}"
-  
   name=$(basename "${dashboard}" | sed -E 's/\.gen\.json$//' | sed -E 's/\.json$//')
   key="${name}.json"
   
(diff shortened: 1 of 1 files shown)


third_party/istio/update-istio.sh
```diff
@@ -52,7 +52,42 @@ cat ${tmpdir}/istio_full.yaml \
...
+  "istio-workload-dashboard.json"
+  "pilot-dashboard.gen.json"
+)
+
+dashboard_yaml="${tmpdir}/dashboards.yaml"
+true > "${dashboard_yaml}"
+
+for dashboard in "${DASHBOARDS[@]}"; do
+  json_file="${tmpdir}/${dashboard}"
+  curl -fsSL "https://raw.githubusercontent.com/istio/istio/${VERSION}/manifests/addons/dashboards/${dashboard}" -o "${json_file}"
```
Tobias-Pe: [CORRECTNESS, PERFORMANCE] Please consider the following updates to improve the dashboard downloads:

* Verify the exit status of the `curl` command to ensure the script fails early if a download fails, preventing invalid YAML generation.
* Optimize the download speed by running them in parallel (using background jobs with `&` and `wait`) or using a batched `curl` invocation to avoid sequential network round-trips.

🤖 **AI Reviewer** (Jetski) | Traces: correctness: go/traj/bb1b5fdf-6a81-49b3-a3b2-5c871244056f, performance: go/traj/233135f6-adce-4d81-ba66-9607b5ea9999

Tobias-Pe: WRT optimiation maybe just a nit or even ignorable imo but the exit code could be relevant

Tobias-Pe (approved):
Thanks !

third_party/istio/update-istio.sh
awhdesmond: Updated script to used batched curl and exit on download failure.

third_party/istio/update-istio.sh
```diff
@@ -52,7 +52,53 @@ cat ${tmpdir}/istio_full.yaml \
...
+DASHBOARDS=(
+  "istio-mesh-dashboard.gen.json"
+  "istio-performance-dashboard.json"
+  "istio-service-dashboard.json"
+  "istio-workload-dashboard.json"
+  "pilot-dashboard.gen.json"
+)
+
+dashboard_yaml="${tmpdir}/dashboards.yaml"
+true > "${dashboard_yaml}"
```
ensonic: Since this is a tempdir, we can probably just use `touch "${dashboard_yaml}"` as there is no need to reset the file.

ensonic (approved):
Thanks!

third_party/istio/update-istio.sh
awhdesmond: Updated!

PR #735, opened 2026-06-18
Switch over nginx-ingress-lb selector to nginx shield

awhdesmond:
Switch over nginx-ingress-lb selector to nginx shield if use_nginx_shield is enabled

NOTE: 30s-1m of downtime for existing setups is expected as the new NGINX shield deployment pods are pulled,  initialized, passing readiness probes and NLB target pools updated.

### DIFF
--- a/src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
+++ b/src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
@@ -1,4 +1,4 @@
-{{- if eq .Values.use_nginx_shield "true" }}
+{{- if and (eq .Values.use_nginx_shield "true") (eq .Values.use_istio "false") }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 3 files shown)


ensonic (approved):
Thanks!

src/app_charts/base/cloud/nginx-ingress-controller-shield.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if eq .Values.use_nginx_shield "true" }}
+{{- if and (eq .Values.use_nginx_shield "true") (eq .Values.use_istio "false") }}
```
awhdesmond, line 1: Be more strict, only deploy shield if `use_nginx_shield=true` and `use_istio=false`

ensonic: Thanks!

ensonic (approved):

PR #737, opened 2026-06-18
Enable gateway api on alternate region GKE clusters

awhdesmond:
Enable gateway api on alternate region GKE clusters


ensonic (approved):
thanks!

PR #738, opened 2026-06-18
Speed up relay tests

ensonic:
* Put relay-broker timeouts into vars that can be shortened by tests
* Use shorter sleeps in to integration test
* Drive by golang fixes (e.g. bad format strings).

### DIFF
--- a/src/go/cmd/http-relay-server/server/broker_test.go
+++ b/src/go/cmd/http-relay-server/server/broker_test.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"os"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


ensonic:
Summary of Changes

   1. Configurable Broker Timeouts:
       * In src/go/cmd/http-relay-server/server/broker.go, I replaced the hardcoded 10 * time.Second and 30 * time.Second timeouts with package-level variables relayTimeout and getRequestTimeout.
       * This allows the tests to override these timeouts without changing the public server.Config API.

   2. Fast Unit Tests:
       * In src/go/cmd/http-relay-server/server/broker_test.go, I added a TestMain function that sets these timeouts to 100ms and 200ms, respectively.
       * This reduced the execution time of //src/go/cmd/http-relay-server/server:go_default_test from ~40 seconds to ~0.5 seconds.

   3. Faster Integration Tests:
       * In src/go/tests/relay/nok8s_relay_test.go, I reduced the polling interval from 1 second to 100 milliseconds when waiting for the relay server and client to start.
       * In src/go/cmd/http-relay-server/server/server.go, I moved the "Relay server listening" log line to occur before Serve, ensuring it's printed as soon as the listener is ready.
       * This reduced the execution time of //src/go/tests/relay:nok8s_relay_test from ~21 seconds to ~10.5 seconds

src/go/cmd/http-relay-server/server/broker_test.go
```diff
@@ -27,6 +28,17 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+func TestMain(m *testing.M) {
+	// Set short timeouts for tests to speed them up.
+	// These can be overridden on a test-by-test basis if needed using:
+	//   oldRelayTimeout := relayTimeout
```
csieber, line 34: We often use this pattern but I wonder if that actually prevents tests running in parallel in the same process.

csieber (approved):

src/go/cmd/http-relay-server/server/broker_test.go
ensonic: Good point. This seems to work. So i'll not over-complicate it. Better would be to pass a config when creating a broker instance.

PR #742, opened 2026-06-19
🧱 base: move GRPCRoute CRD to general installation

Tobias-Pe:
Extract the GRPCRoute CRD from the Istio-only manifests and package it as a general cloud CRD in the base-cloud chart.

#### Why

Throughout the migration of ingress to gateway, we want to already submit verified & working http and grpcroutes into our vcs on intrinsic's private repository. But our release would fail on deploying the inactive grpcroutes if the cluster is still using ingress. To support deploying grpcroutes on a cluster which doesn't use them, to e.g. allow killswitch OR A/B deployments we add the grpcroute (like the httproute that is installed by default to GKE) into all clusters.

- Created a new `third_party/gateway-api` package with an `update-gateway.sh` script to download and manage the `GRPCRoute` CRD.
- Removed the `GRPCRoute` CRD download and prepending logic from `third_party/istio/update-istio.sh` and regenerated `istio-generated.yaml` without it.
- Added the `GRPCRoute` CRD (`grpcroute-generated.yaml`) to the `base-cloud` chart's `extra_templates` in `src/app_charts/base/BUILD.bazel` so it is always installed on the cloud.

go/traj/811a94be-1b2e-490a-aaf0-df6e86fbd0ee


awhdesmond (approved):

awhdesmond:
There are some re-ordering of the lines within the files, but I suppose those should not have any impact.

ensonic (approved):

ensonic:
> There are some re-ordering of the lines within the files, but I suppose those should not have any impact.

Looks like we need some bazel formatting presubmit check as well.

PR #744, opened 2026-06-19
bump nginx shield version; increase resource limits

awhdesmond:
bump nginx shield version;  https://my.f5.com/manage/s/article/K000161614

increase resource limits to allow for larger load


csieber (approved):

ensonic (approved):
Thanks

PR #755, opened 2026-06-25
Add a benchmark for the relay.

ensonic:
For this we had to refactor the relay to add context passing and cancellation handling in a few places.

### DIFF
--- a/src/go/cmd/http-relay-client/client/client.go
+++ b/src/go/cmd/http-relay-client/client/client.go
@@ -155,7 +155,7 @@ func NewClient(config ClientConfig) *Client {
 	return c
 }
 
-func (c *Client) Start() {
+func (c *Client) Start(ctx context.Context) {
 	var err error
 
 	remoteTransport := http.DefaultTransport.(*http.Transport).Clone()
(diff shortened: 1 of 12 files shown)


src/go/cmd/http-relay-client/client/client.go
```diff
@@ -254,9 +254,17 @@ func (c *Client) Start() {
 	wg := new(sync.WaitGroup)
 	wg.Add(c.config.NumPendingRequests)
 	for i := 0; i < c.config.NumPendingRequests; i++ {
-		go c.localProxyWorker(remote, local)
-	}
-	// Waiting for all goroutines to finish (they never do)
+		go func() {
+			defer wg.Done()
+			c.localProxyWorker(ctx, remote, local)
```
csieber, line 257: Can we just use waitgroup.Go here?

csieber (approved):
cool!

src/go/cmd/http-relay-client/client/client.go
ensonic: Absolutely. Done.

PR #756, opened 2026-06-26
Add config to set vm type for gke nodes

koonpeng:


scripts/include-config.sh
```diff
@@ -62,4 +62,8 @@ function include_config {
 
   GKE_DATAPATH_PROVIDER=${GKE_DATAPATH_PROVIDER:-DATAPATH_PROVIDER_UNSPECIFIED}
   check_var_is_one_of GKE_DATAPATH_PROVIDER "DATAPATH_PROVIDER_UNSPECIFIED" "ADVANCED_DATAPATH"
+
+  GCP_VM_TYPE=${GCP_VM_TYPE:-e2-standard-4}
+  GKE_MIN_NODES=${GKE_MIN_NODES:-1}
```
ensonic: since initial node count is 2, I wonder if we ever need to scale below 2?

scripts/include-config.sh
```diff
@@ -62,4 +62,8 @@ function include_config {
 
   GKE_DATAPATH_PROVIDER=${GKE_DATAPATH_PROVIDER:-DATAPATH_PROVIDER_UNSPECIFIED}
   check_var_is_one_of GKE_DATAPATH_PROVIDER "DATAPATH_PROVIDER_UNSPECIFIED" "ADVANCED_DATAPATH"
+
+  GCP_VM_TYPE=${GCP_VM_TYPE:-e2-standard-4}
```
ensonic: GKE_NODE_VM_TYPE

scripts/include-config.sh
koonpeng: done

scripts/include-config.sh
koonpeng: done

ensonic (approved):

src/bootstrap/cloud/terraform/cluster.tf
```diff
@@ -4,6 +4,10 @@
 # service account for the nodes. This service account cannot be used by the
 # workloads: see workload-identity.tf for those service accounts.
 
+locals {
+  node_disk_type = var.node_disk_type != null ? var.node_disk_type : (startswith(var.node_machine_type, "n4") ? "hyperdisk-balanced" : "pd-balanced")
+}
+
```
ensonic, line 10: Is that meant to be part of this PR?

drigz (approved):

PR #758, opened 2026-06-26
Cleanup context handling

ensonic:
1. Modernized Graceful Shutdown: Updated all entry-point main.go files to use os/signal.NotifyContext. This provides a base context that is automatically cancelled upon receiving termination signals (SIGINT/SIGTERM), allowing for graceful cleanup of all downstream components.
   - Updated: setup-dev, setup-robot, token-vendor, cr-syncer, http-relay-client, metadata-server, and gcr-credential-refresher.
2. Context Propagation in Controllers: Refactored the Kubernetes controllers (chart-assignment-controller and app-rollout-controller) to use signals.SetupSignalHandler() for their base context, which is then passed down to reconcilers and background release managers.
3. Refactored Library Functions: Updated configutil.ReadConfig and Synk.Init to accept a context.Context as their first parameter. This ensures that these long-running or network-dependent operations can be cancelled by their callers.
4. Eliminated Deprecated Contexts: Replaced the obsolete golang.org/x/net/context with the standard library's context package in cr-syncer.
5. Audit of context.TODO() and context.Background():
   - Confirmed that context.TODO() is not used (only in generated code).
   - Reduced the usage of context.Background() in production code, replacing it with passed-down contexts to ensure proper cancellation propagation.
6. Cleanup of tests to use t.Context()

### DIFF
--- a/src/go/cmd/metadata-server/main.go
+++ b/src/go/cmd/metadata-server/main.go
@@ -95,7 +95,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctx := context.Background()
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
 
 	config, err := rest.InClusterConfig()
 	if err != nil {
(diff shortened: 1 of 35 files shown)


ensonic:
Sorry for the large PR, the majority is using t.Context() in tests though.

src/go/cmd/metadata-server/main.go
```diff
@@ -188,12 +189,9 @@ func main() {
...
 	}()
 
-	stop := make(chan os.Signal)
-	signal.Notify(stop, syscall.SIGTERM)
-
-	<-stop
+	<-ctx.Done()
 	if !*runningOnGKE {
-		RevertCorefile(ctx, k8s)
+		RevertCorefile(context.Background(), k8s)
```
csieber, line 195: Can you add a comment why Background is used it? In the first moment this is surprising.

csieber (approved):

src/go/cmd/metadata-server/main.go
ensonic: Done. I would like to refactor the code, but I don't want to add even more changes here.

PR #759, opened 2026-06-26
Update the defaults for the relay.

ensonic:
Add comment to explain the chosen values. Replace the TODO in the relay client with an explanation why we need the chunking.


csieber (approved):

hheibel (commented):
Thanks Stefan. This looks generally good. Do we have a way to test these changes? Will this happen via existing unit or post-submit tests or will this happen by exercising the relay on intrinsic-dev or similar projects?

ensonic:
> Thanks Stefan. This looks generally good. Do we have a way to test these changes? Will this happen via existing unit or post-submit tests or will this happen by exercising the relay on intrinsic-dev or similar projects?

The tests are both unit and integration tested.

hheibel (approved):

PR #760, opened 2026-06-29
Fix memory leak in http-relay-server.

ensonic:
When a request is stopped (e.g., if the user-client times out or closes the connection), StopRelayRequest was simply deleting the request from the internal map. However, it was not closing the associated responseStream and requestStream channels. This caused the goroutines ranging over these channels (started in Server.responseFilter and Server.bidirectionalStream) to block indefinitely, leading to a steady increase in goroutines and memory usage over time.

This change refactores the cleanup logic into a common removeRequest helper that explicitly closes all channels and signals the goroutines to exit before removing the request from the map. It ensures that resources are always reclaimed regardless of whether the request timed out or was explicitly stopped.

In addition this adds a new test to cover this and pprof support (so that we can easily identify similar cases in the future).

### DIFF
--- a/src/go/cmd/http-relay-server/main.go
+++ b/src/go/cmd/http-relay-server/main.go
@@ -25,7 +25,10 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"log/slog"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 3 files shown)


ensonic:
Tested pprof via:
```
kubectl --context=<myctx> port-forward -n app-prometheus deployment/prometheus-relay-server 7654:8765
```

drigz (approved):

drigz:
Nice find, thank you!

PR #765, opened 2026-06-30
Replace gomega with regular go test mechanisms.

ensonic:
This was only used in 3 tests and while it is more verbose, it simplifies dependency management and aligns the tests.


ensonic:
We can also go the other way around. That is keep the dependency and use it in more tests. Any opinion?

drigz (approved):
new tests seem simpler to me

PR #766, opened 2026-06-30
Add more tests to drive up coverage.

ensonic:
Main goal is to get more peace of mind when accepting dependabot changes.


ensonic:
@therjak Sorry for the large PR. 

I am not entirely happy with the controller_tests.go yet. I'll send a followup to split the approllout/validator like we do for chart assignments.

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -856,3 +867,430 @@ metadata:
...
+			}
+
+			if tc.verify != nil {
+				tc.verify(t, client)
+			}
+		})
+	}
+}
+
+func ptrBool(b bool) *bool {
```
therjak: the go.mod claims the code needs go 1.26. so you should be able to use code like 'new(true)' instead. https://go.dev/doc/go1.26#language

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -856,3 +867,430 @@ metadata:
 		})
 	}
 }
+
+func TestReconciler_Reconcile(t *testing.T) {
+	scheme := runtime.NewScheme()
```
therjak, line 872: fyi: this block seems to modify global state so the tests can not be run in parallel.

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -856,3 +867,430 @@ metadata:
...
+					Object: runtime.RawExtension{Raw: []byte("invalid json")},
+				},
+			},
+			wantAllow: false,
+		},
+		{
+			name: "Allowed",
+			req: admission.Request{
+				AdmissionRequest: admissionv1.AdmissionRequest{
+					Object: runtime.RawExtension{Raw: arValidRaw},
```
therjak: having the value so far from the test case makes the code more difficult to read. is it just to skip the error value of json.Marshal? you could just add a small wrapper that does not return the error.

src/go/pkg/controller/approllout/controller_test.go
```diff
@@ -856,3 +867,430 @@ metadata:
...
+	ca := &apps.ChartAssignment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "my-ca",
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "apps.cloudrobotics.com/v1alpha1",
+					Kind:       "AppRollout",
+					Name:       "owner-rollout",
+					UID:        "owner-uid",
+					Controller: &_true,
```
therjak: please replace with new(true)

src/go/pkg/gcr/update_gcr_credential_test.go
```diff
@@ -42,3 +52,198 @@ func TestDockercfgJSON(t *testing.T) {
...
+				return
+			}
+
+			if tt.wantErr {
+				return
+			}
+
+			// Verify secrets that should exist and be updated
+			for _, ns := range tt.wantSecretNS {
+				secret, err := fakeK8s.CoreV1().Secrets(ns).Get(context.Background(), SecretName, metav1.GetOptions{})
```
therjak: why not t.Context()? below as well

src/go/pkg/controller/chartassignment/validator_test.go
```diff
@@ -145,3 +152,90 @@ spec:
 		})
 	}
 }
+
+func TestHandle(t *testing.T) {
+	scheme := runtime.NewScheme()
+	if err := apps.AddToScheme(scheme); err != nil {
+		t.Fatalf("failed to add apps to scheme: %v", err)
```
therjak, line 158: why are you testing here but not all the other new apps.AddToScheme?

src/go/pkg/controller/chartassignment/validator_test.go
```diff
@@ -145,3 +152,90 @@ spec:
...
+				},
+			},
+			wantAllowed: false,
+			wantCode:    http.StatusBadRequest,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resp := v.Handle(context.Background(), tt.req)
```
therjak: t.Context()

src/go/pkg/controller/chartassignment/release_test.go
```diff
@@ -126,3 +146,240 @@ spec:
...
 }
+
+func pollStatus(t *testing.T, rs *releases, name string, check func(releaseStatus) bool) {
+	t.Helper()
+	for i := 0; i < 100; i++ {
+		status, ok := rs.status(name)
+		if ok && check(status) {
+			return
+		}
+		time.Sleep(10 * time.Millisecond)
```
therjak: if you are using time.Sleep inside a test, please use them within https://pkg.go.dev/testing/synctest

src/go/pkg/controller/chartassignment/release_test.go
```diff
@@ -126,3 +146,240 @@ spec:
...
+
+func TestReleases_EnsureUpdated_Busy(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockSynk := NewMockInterface(ctrl)
+	recorder := &record.FakeRecorder{
+		Events: make(chan string, 10),
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
therjak: t.Context

src/go/pkg/controller/chartassignment/release_test.go
```diff
@@ -126,3 +146,240 @@ spec:
...
+
+func TestReleases_EnsureUpdated_TransientErrorRetry(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockSynk := NewMockInterface(ctrl)
+	recorder := &record.FakeRecorder{
+		Events: make(chan string, 10),
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
therjak: t.context

src/go/pkg/controller/chartassignment/release_test.go
```diff
@@ -126,3 +146,240 @@ spec:
...
+
+func TestReleases_EnsureUpdated_PermanentErrorNoRetry(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockSynk := NewMockInterface(ctrl)
+	recorder := &record.FakeRecorder{
+		Events: make(chan string, 10),
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
therjak: t.context

src/go/pkg/controller/chartassignment/release_test.go
```diff
@@ -126,3 +146,240 @@ spec:
...
+
+func TestReleases_EnsureUpdated(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockSynk := NewMockInterface(ctrl)
+	recorder := &record.FakeRecorder{
+		Events: make(chan string, 10),
+	}
+	ctx, cancel := context.WithCancel(context.Background())
```
therjak: t.context and you probably do not need to cancel as t.context gets canceled at the end of the test

therjak (changes requested):

src/go/pkg/controller/approllout/controller_test.go
ensonic: Nice. This indeed works. Did not know about it.

src/go/pkg/controller/approllout/controller_test.go
ensonic: good idea. Done

src/go/pkg/gcr/update_gcr_credential_test.go
ensonic: Totally. Changed it in all of the tests.

src/go/pkg/controller/chartassignment/release_test.go
ensonic: done

src/go/pkg/controller/chartassignment/release_test.go
ensonic: Done. This is much better, as it triggered improving the error handling on the release.go side :)

src/go/pkg/controller/approllout/controller_test.go
ensonic: scheme is a local var. Where do you see the global state?

ensonic:
100 thanks for the in-depth review.

src/go/cmd/http-relay-server/server/broker_test.go
```diff
@@ -270,98 +271,100 @@ func TestTimeout(t *testing.T) {
...
+		wg.Add(1)
+		go func() {
+			reqErr = b.SendResponse(&pb.HttpResponse{Id: req.Id, Body: []byte(*req.Id), Eof: proto.Bool(false)})
+			if !errors.Is(reqErr, ErrClosedInactivity) {
+				t.Errorf("Wrong SendResponse error: %v; want %v", reqErr, ErrClosedInactivity)
+			}
+			wg.Done()
+		}()
+		// Wait for the goroutine to be blocked on writing the response.
+		synctest.Wait()
```
koonpeng, line 315: Learned something new 👍

src/go/pkg/controller/approllout/controller_test.go
therjak: I was referring to apps.AddToScheme, registry.AddToScheme, core.AddToScheme. they store in package level global vars.

therjak (approved):

src/go/pkg/controller/approllout/controller_test.go
ensonic: I think they add the changes to 'schema'. But it ready weirdly.

ensonic: Thanks. I'll ignore this for now. Gemeni don't think there is local state involved. The code goes deep into the k8s and I am hitting deprecated apis. I'll explore this another day.

PR #768, opened 2026-07-01
move go.mod from src into the repository root

therjak:
having it in a subdirectory seems to confuse some tooling. follow most other projects and place it in the root.
this will probably confuse the next update of dependent projects. but as dependabot it among the confused tooling they seem to require a manual update anyway.


BUILD.bazel
```diff
@@ -8,6 +8,8 @@ package(default_visibility = ["//visibility:public"])
 exports_files([
     "config.sh.tmpl",
     "deploy.sh",
+    "go.mod",
+    "go.sum",
```
ensonic: do we need to export go.sum?

ensonic (approved):

BUILD.bazel
therjak: looks like we do not. removed the export.

PR #774, opened 2026-07-03
Update kube-prometheus-stack from 72.9.1 to 87.5.1 (appVersion v0.92.1)

ensonic:


ensonic:
I've deployed this to my test project, and it runs, but can't access it (while testing the istio setup). Using port-forwards failed due to the path prefix rewrite we do. I need to think about testing instructions, let me know if you have some ideas.

drigz (approved):
You can deploy to my test project if you want - or just merge and we test the daily build on the shared test project next week.

I never figured out a way to connect via port-forward with the rewrite, but I think Markus might have figured it out (maybe temporarily disabling the base URL by editing the values that configure it??)

ensonic:
I'll wait until monday.

PR #776, opened 2026-07-03
We need to double the oauth2 proxy routes

ensonic:
When  extAuthz  finishes its check against  crc-auth-gateway  and approves a request by returning 200 OK (for instance, when calling /oauth2/userinfo with a valid session cookie, or accessing an authenticated endpoint under /web-apis ), Envoy at crc-gateway proceeds to forward the actual request.

oauth2-proxy-interactive  (/oauth2) and  oauth2-proxy (/web-apis): Tell crc-gateway where to send the physical HTTP request once authorization has succeeded. Without these routes on crc-gateway, approved requests would result in a  404 Not Found  at the public gateway layer.

### DIFF
--- a/src/app_charts/base/cloud/oauth2-proxy.yaml
+++ b/src/app_charts/base/cloud/oauth2-proxy.yaml
@@ -100,17 +100,19 @@ spec:
             port:
               name: http
 ---
+{{- range list "" "-auth" }}
+{{- $sectionName := "https" }}{{ if eq . "-auth" }}{{ $sectionName = "http" }}{{ end }}
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
-  name: oauth2-proxy
+  name: oauth2-proxy{{ . }}
 spec:
   hostnames:
-  - {{ .Values.domain }}
+  - {{ $.Values.domain }}
   parentRefs:
-  - name: crc-gateway
+  - name: crc{{ . }}-gateway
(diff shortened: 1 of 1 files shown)


src/app_charts/base/cloud/oauth2-proxy.yaml
```diff
@@ -100,17 +100,20 @@ spec:
             port:
               name: http
 ---
+{{- range list "" "-auth" }}
+{{- $sectionName := "https" }}{{ if eq . "-auth" }}{{ $sectionName = "http" }}{{ end }}
+{{- $gatewayName := "crc-gateway" }}{{ if eq . "-auth" }}{{ $gatewayName = "crc-auth-gateway" }}{{ end }}
```
Tobias-Pe: [STYLE] Prefer using `name: crc{{ . }}-gateway` directly for parentRefs instead of defining `$gatewayName`, to simplify the template and maintain consistency with other route definitions across charts (such as in `kubernetes-api.yaml` and `grafana-http-route.yaml`).

🤖 **AI Reviewer** (Jetski) | Traces: style: http://go/traj/2834551d-525e-41fc-9de1-0ac01eb31f82

Tobias-Pe (approved):
👍

src/app_charts/base/cloud/oauth2-proxy.yaml
ensonic: done.

PR #780, opened 2026-07-06
EnvoyFilter to perform auth-signin

awhdesmond:
EnvoyFilter to replicate NGINX auth-signin functionality

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -214,7 +214,7 @@ spec:
   targetRefs:
     - group: gateway.networking.k8s.io
       kind: Gateway
-      name: crc-gateway-auth
+      name: crc-auth-gateway
   configPatches:
   - applyTo: HTTP_FILTER
     match:
@@ -251,6 +251,9 @@ spec:
             end
 
             function envoy_on_request(request_handle)
+              -- Remove the internal redirect header so it doesn't leak to the client
+              request_handle:headers():remove(X_AUTH_SIGNIN_H)
+
               local path = request_handle:headers():get(":path")
               request_handle:streamInfo():dynamicMetadata():set(DYNAMIC_METADATA_NAMESPACE, "request_path", path)
             end
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -133,14 +133,14 @@ spec:
             path: /readyz
             port: 8080
             scheme: HTTP
-          initialDelaySeconds: 2 
+          initialDelaySeconds: 2
           periodSeconds: 5
         livenessProbe:
-          httpGet: 
+          httpGet:
             path: /healthz
             port: 8080
             scheme: HTTP
-          initialDelaySeconds: 2 
+          initialDelaySeconds: 2
           periodSeconds: 5
         resources:
           requests:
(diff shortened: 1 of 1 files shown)


src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,95 @@ spec:
...
+      value:
+        name: envoy.lua.crc_auth_signin
+        typed_config:
+          "@type": "type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua"
+          inlineCode: |
+            local STATUS_302 = "302"
+            local LOCATION_H = "location"
+            local X_AUTH_SIGNIN_H = "x-crc-auth-signin"
+
+            local DEFAULT_SIGNIN_URL = {{ .Values.signin_url }}
```
awhdesmond: The default auth-signin URL. It can be removed if we don't want to redirect any 401/403 by default

src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,95 @@ spec:
...
+                return str
+              end
+
+              str = string.gsub(str, "([^A-Za-z0-9_.-])", function(c)
+                return string.format("%%%02X", string.byte(c))
+              end)
+              return str
+            end
+
+            function envoy_on_request(request_handle)
```
awhdesmond, line 253: Need to track `request_path` when handing the request. Will be used later during response handling.

src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,85 @@ spec:
...
+      value:
+        name: envoy.lua.crc_auth_signin
+        typed_config:
+          "@type": "type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua"
+          inlineCode: |
+            local STATUS_302 = "302"
+            local LOCATION_H = "location"
+            local X_AUTH_SIGNIN_H = "x-crc-auth-signin"
+
+            local DEFAULT_SIGNIN_URL = "{{ .Values.signin_url }}"
```
awhdesmond, line 239: DEFAULT_REDIRECT_URL if no auth-signin header is found.

src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,85 @@ spec:
...
+
+              -- Read and remove the internal redirect header so it doesn't leak to the client
+              local auth_signin_url = response_handle:headers():get(X_AUTH_SIGNIN_H)
+              if auth_signin_url then
+                response_handle:headers():remove(X_AUTH_SIGNIN_H)
+              end
+
+              if not auth_signin_url or auth_signin_url == "" then
+                auth_signin_url = DEFAULT_SIGNIN_URL
+              end
```
awhdesmond, line 278: If we do not want a default redirect, we can also just return here. Which means, the presence of the header will be used to control if redirect should be performed.

src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,85 @@ spec:
...
+
+              -- Substitute $escaped_request_uri in the redirect URL
+              local meta = response_handle:streamInfo():dynamicMetadata():get(DYNAMIC_METADATA_NAMESPACE)
+              local request_path = meta and meta["request_path"]
+              if request_path then
+                local escaped_path = url_encode(request_path)
+                auth_signin_url = string.gsub(auth_signin_url, "%$escaped_request_uri", function() return escaped_path end)
```
awhdesmond, line 285: replace `$escaped_request_uri` with our `request_path` saved earlier.

src/app_charts/base/cloud/istio.yaml
```diff
@@ -204,4 +204,85 @@ spec:
...
+              local status = response_handle:headers():get(":status")
+
+              if status ~= "403" and status ~= "401" then
+                -- Only redirect 401 (Unauthorized) and 403 (Forbidden) responses;
+                -- let all other statuses pass through.
+                return
+              end
+
+              -- Read and remove the internal redirect header so it doesn't leak to the client
+              local auth_signin_url = response_handle:headers():get(X_AUTH_SIGNIN_H)
```
Tobias-Pe, line 271: We are potentially opening a Open Redirect vulnerability here. 

E.g. attack vector:

If a malicious actor wants to execute a phishing attack, they might try to manipulate the redirect URL. The attack flows like this:

The Injection: An attacker crafts a malicious link and tricks a legitimate user into clicking it. When the user's browser makes the request to your Gateway, the attacker ensures a custom header is included in the request:
x-crc-auth-signin: [https://evil-phishing-site.com](https://evil-phishing-site.com)

The Echo (The Vulnerability): The request reaches your ExtAuthz service or backend app. If that service is misconfigured or simply echoes unknown headers back in its 401/403 response, it will send x-crc-auth-signin: [https://evil-phishing-site.com](https://evil-phishing-site.com) back to Envoy.

The Execution: The Lua script catches the 401/403 response, reads the injected header, and dutifully constructs a 302 Found redirecting the user to [https://evil-phishing-site.com/?rd=](https://evil-phishing-site.com/?rd=)....

The Impact: The user thinks they are being redirected to your standard login page, but they land on a fake page controlled by the attacker, leading to stolen credentials.


Can we secure somehow against this ? Only allow a list of targets ?

Tobias-Pe (changes requested):

src/app_charts/base/cloud/istio.yaml
Tobias-Pe: The easiest fix could be to add to function envoy_on_request(request_handle)

              request_handle:headers():remove("x-crc-auth-signin")

WDYT ?

awhdesmond: added!

Tobias-Pe (commented):
Could you add instructions on how we expect a service to supply the header ? 

Is something like this expected / wanted ? : 

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: protected-route
spec:
  parentRefs:
  - name: my-gateway
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /secure
    filters:
    - type: ResponseHeaderModifier
      responseHeaderModifier:
        set:
        - name: x-crc-auth-signin
          value: "https://auth.example.com/login?rd=$escaped_request_uri"
    backendRefs:
    - name: my-backend-service
      port: 8080
```


Then we would need to remove the header then in your PR also on 200 successfull to prevent:
The Catch: Header Leaking
Because an HTTPRoute cannot conditionally apply filters based on response codes, it will attach this header to EVERY response from my-backend-service—including successful 200 OK responses.

awhdesmond:
> Could you add instructions on how we expect a service to supply the header ?
> 
> Is something like this expected / wanted ? :
> 
> ```yaml
> apiVersion: gateway.networking.k8s.io/v1
> kind: HTTPRoute
> metadata:
>   name: protected-route
> spec:
>   parentRefs:
>   - name: my-gateway
>   rules:
>   - matches:
>     - path:
>         type: PathPrefix
>         value: /secure
>     filters:
>     - type: ResponseHeaderModifier
>       responseHeaderModifier:
>         set:
>         - name: x-crc-auth-signin
>           value: "https://auth.example.com/login?rd=$escaped_request_uri"
>     backendRefs:
>     - name: my-backend-service
>       port: 8080
> ```
> 
> Then we would need to remove the header then in your PR also on 200 successfull to prevent: The Catch: Header Leaking Because an HTTPRoute cannot conditionally apply filters based on response codes, it will attach this header to EVERY response from my-backend-service—including successful 200 OK responses.

This is an example to demonstrate how it could be done, I am not entirely sure if header-based auth-signin is required. If DEFAULT_SIGNIN_URL is sufficient, we will not need header-based auth-signin?

awhdesmond:
Or another possible approach would be to use the head as a flag: `x-crc-auth-signin: true` and redirect the user to the default_signin_url.

PR #782, opened 2026-07-06
Add a new annotation to temporarily bypass synk.

ensonic:
This allows for a `kubectl edit` based workflow to make temporary changes, e.g. enable memory profiling.

### DIFF
--- a/src/go/pkg/synk/synk.go
+++ b/src/go/pkg/synk/synk.go
@@ -57,6 +57,11 @@ import (
 // src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go
 const totalAnnotationSizeLimitB int = 256 * (1 << 10) // 256 kB
 
+// Annotation to opt-out of resource synchronization. If set to "true" on a resource
+// in the cluster, Synk will not overwrite it with the chart version, but will still
+// update its owner references so it does not get pruned.
+const AnnotationIgnore = "synk.cloudrobotics.com/ignore"
+
 // Synk allows to synchronize sets of resources with a fixed cluster.
 type Synk struct {
 	discovery   discovery.CachedDiscoveryInterface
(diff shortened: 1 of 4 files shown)


src/go/pkg/synk/synk.go
```diff
@@ -727,6 +732,18 @@ func (s *Synk) applyOne(ctx context.Context, resource *unstructured.Unstructured
 		return apps.ResourceActionNone, fmt.Errorf("owner conflict: %w", err)
 	}
 
+	if current.GetAnnotations()[AnnotationIgnore] == "true" {
+		setOwnerRef(current, set)
```
therjak, line 742: where does setOwnerRef get called without the annotation? I only see a comment in line 800. shouldn't is get updated not matter the annotation?

ensonic: Added a comment to clarify. We're already calling setOwnerRef in applyAll. But here we're reverting to the current version.

Would it be easier to read if we return some extra flag and do the change one level up?

therjak: it would be ugly as well but it would prevent inconsistent behavior. I see 3 places calling 'applyOne' but only for one I directly see the setOwnerRef call. (another passes nil as owner, so that one hopefully just does nothing and not remove an owner). but the first one in applyAll is unclear.

csieber (approved):

src/go/pkg/synk/synk.go
ensonic: The call chain is Apply -> applyAll -> applyOne

* applyAll calls setOwnerRef to point it to the new one
* applyOne calls setOwnerRef to revert that in thes code path

Options:
1) submit as it is
2) move setOwnerRef() call from applyAll to applyOne
3) return something from applyOne and move the special case setOwnerRef call to applyAll

I think given the comment 1) is okay. 2) might also be okay. 3) is ugly.

ensonic: I'd leave the asymmetry for now, but I've handled all cases. I'll see if I can refactor the code next. If you prefer, I can do (2) .

therjak (approved):

PR #784, opened 2026-07-06
Fix namespace ref for token-vendor.

ensonic:
Add missing namespace + required reference grant.


Tobias-Pe (approved):
Nice ! Saw the error already and you just fixed it :D

PR #785, opened 2026-07-06
🧱 base-cloud: Comment out hostname in external gateway listeners

Tobias-Pe:
Comment out the hostname field in http and https listeners of the external crc-gateway.
This allows TLS handshakes without SNI by default, fixing issues with clients/proxies that do not send SNI.
Added comments explaining how to uncomment if host-based routing is needed (e.g. if multiple gateways share the same IP).

#### Why

Allow TLS handshakes without SNI by default to support more clients, while providing instructions to enable it if needed for multi-gateway setups.

- Comment out hostname in http listener in base-cloud istio.yaml template.
- Comment out hostname in https listener in base-cloud istio.yaml template.

go/traj/7de9d91c-7964-436e-9be8-f6d4f8a308ab
TESTED=Manually tested
RELNOTES=NONE
TAG=agy
CONV=13a705e3-635c-4db8-aa9e-0ef4d8f0e30d

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -14,14 +14,14 @@ spec:
     type: IPAddress
   listeners:
   - name: http
-    hostname: {{ .Values.domain }}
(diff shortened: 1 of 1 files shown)


Tobias-Pe:

ensonic (approved):

PR #786, opened 2026-07-07
🐛 cr-syncer-auth-webhook: fallback to X-Envoy-Original-Path under path rewrites

Tobias-Pe:
Update validateRequest in cr-syncer-auth-webhook to check X-Envoy-Original-Path when X-Original-Url is empty.

#### Why

When cr-syncer-auth-webhook is deployed behind an Envoy-based Kubernetes Gateway (such as Istio or Gateway API with URLRewrite filters), Envoy strips `X-Original-Url` and preserves the original request path in `X-Envoy-Original-Path`. Because `cr-syncer-auth-webhook` checked only `X-Original-Url`, requests rewritten by Envoy resulted in an empty path (`""`) and were rejected with 400 Bad Request during certificate synchronization checks.

- Introduce `extractOriginalURL` helper function in `main.go` to fall back to `X-Envoy-Original-Path` when `X-Original-Url` is empty.
- Update log message to `unexpected value of origin URL header` when URL parsing fails.

### DIFF
--- a/src/go/cmd/cr-syncer-auth-webhook/main.go
+++ b/src/go/cmd/cr-syncer-auth-webhook/main.go
@@ -109,10 +109,11 @@ func (h *handlers) resourceIsFiltered(groupKind string) bool {
 // validateRequest checks that the request is expected for the cr-syncer and
 // only accesses allowed resources.
 func (h *handlers) validateRequest(r *http.Request, robotName string) error {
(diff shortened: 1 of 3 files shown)


Tobias-Pe:

ensonic (approved):
Thanks!

PR #788, opened 2026-07-07
🐛 token-vendor: add bypass route on crc-auth-gateway for verify/oauth2 endpoints

Tobias-Pe:
Add token-vendor-auth-bypass HTTPRoute on crc-auth-gateway routing to status-echo (200 OK) to prevent 404 responses during external authorization checks.

#### Why

When external authorization (`ext_authz`) verifies requests to `/apis/core.token-vendor/v1/token.oauth2`, `/token.verify`, or `/jwt.verify`, `crc-auth-gateway` previously lacked a matching route and fell back to `crc-auth-catch-all` (`404 Not Found`). Because `token-vendor` handles its own authentication for these endpoints, we add an `HTTPRoute` on `crc-auth-gateway` targeting `status-echo:80` (`200 OK`) so `crc-gateway` permits the request to reach the `token-vendor` service.

- Add `HTTPRoute/token-vendor-auth-bypass` right between `token-vendor` and `token-vendor-grant` in `src/app_charts/token-vendor/cloud/http-route.yaml`.

### DIFF
--- a/src/app_charts/base/cloud/istio.yaml
+++ b/src/app_charts/base/cloud/istio.yaml
@@ -204,4 +204,18 @@ spec:
     targetPort: 8080
   selector:
     app: crc-auth-gateway-catchall-svc
+---
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: ReferenceGrant
+metadata:
+  name: catchall-svc-grant
+spec:
+  from:
+  - group: gateway.networking.k8s.io
+    kind: HTTPRoute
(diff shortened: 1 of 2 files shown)


Tobias-Pe:

ensonic (approved):
Thanks

Tobias-Pe:
@ensonic i fixed the namespaces for the new resources as they were mixed PTAL

ensonic (approved):

PR #789, opened 2026-07-08
Address continous memory growth in the http-relay server.

ensonic:
The  pprof  heap profile confirmed that a large amount of memory was held by Prometheus metrics (histograms and counters). These metrics use the backend label (which corresponds to the robot name). In integration testing environments where robots are frequently created and deleted, the number of unique backend label values grows indefinitely. Since Prometheus metrics are never automatically cleaned up, this causes a persistent memory leak.

This change implementes a cleanup mechanism for inactive backends:
1. Activity Tracking: The  broker  now tracks the  lastActivity  time for each backend (in broker.go).
1. Periodic Cleanup: The server now periodically (every 10 seconds) calls ReapInactiveBackends  (in server.go).
1. Metrics Deletion: If a backend has been inactive for more than 30 minutes (configurable via flag), it is removed from the broker's internal map, and all its associated Prometheus metrics are deleted from the registry.


drigz (approved):
Thanks!

PR #791, opened 2026-07-08
Add support for x-crc-tv-robots header.

csieber:
`HTTPRoute` `replaceFullPath` does not support query arguments. This change adds support for the header `x-crc-tv-robots` to the token-vendor. See b/532472881 for details.


src/app_charts/k8s-relay/cloud/http-route.yaml
```diff
@@ -57,11 +57,16 @@ spec:
         value: /apis/core.kubernetes-relay/server
 {{- if eq . "-auth" }}
     filters:
+    - type: RequestHeaderModifier
+      requestHeaderModifier:
+        set:
+        - name: x-crc-tv-robots
```
Tobias-Pe, line 63: I think we need to add this header now to allowed to be forwarded by istio

Tobias-Pe (changes requested):

src/app_charts/k8s-relay/cloud/http-route.yaml
csieber: Shouldn't this header be set after any fowarding is done?

Tobias-Pe: Riiight, we only need the whitelist if its a response header. Then i think its okay

src/app_charts/prometheus/values-cloud.yaml
```diff
@@ -40,10 +40,16 @@ gf_server_root_url: "https://${CLOUD_ROBOTICS_DOMAIN}/grafana"
 gf_csrf_trusted_origins: ""
 # Grafana Ingress configuration. Does not use the same replace as the values above.
 gf_ingress_auth_url: "http://oauth2-proxy.default.svc.cluster.local/apis/core.token-vendor/v1/token.verify"
+# Header modification for Istio/HTTPRoutes support.
+gf_ingress_auth_url_header: ""
+gf_ingress_auth_url_headervalue: ""
```
Tobias-Pe, line 45: let's rather introduce the new header which is for the istio/gateway under gateway name instead of ingress WDYT ?

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -381,8 +382,11 @@ func (h *HandlerContext) verifyTokenHandler(w http.ResponseWriter, r *http.Reque
 	w.Write([]byte("OK"))
 }
 
-// testForRobotACL determines if the "robots" parameter is set.
-func testForRobotACL(u *url.URL) bool {
+// testForRobotACL determines if the "robots" parameter or "X-CRC-TV-ROBOTS" header is set.
+func testForRobotACL(u *url.URL, h *http.Header) bool {
+	if headerVal := h.Get(HeaderRobots); headerVal != "" {
+		return strings.ToLower(headerVal) == "true"
```
Tobias-Pe, line 389: in docs there is little mentioning about prio of query vs header, imo we should mention that this has high prio

src/app_charts/prometheus/values-cloud.yaml
csieber: We should also have to replace `gf_ingress_auth_url` with `gf_gateway_auth_url`.  For now I think we should keep `gf_ingress_*` and just redefine it as just "incoming traffic". "ingress traffic" is a common networking term.

src/go/cmd/token-vendor/api/v1/v1.go
csieber: done

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -36,6 +36,7 @@ import (
 )
 
 const (
+	HeaderRobots  = "x-crc-tv-robots"
```
ensonic: does it need to be public?

ensonic (approved):
Thanks

src/go/cmd/token-vendor/api/v1/v1.go
csieber: done

Tobias-Pe (approved):
Thanks!

LGTM GIF

PR #796, opened 2026-07-09
Remove the private CA feature.

ensonic:
This never went to a stable state. The certs it would provide have a private-ca as a root. That makes them less useful in the wild.

Hence manually revert the changes from 1c2898b  (PR #122) to simplify our setup.


ensonic:
Tested on my personal dev project. Terraform does not destroy anything as the resources where only conditionally created. We'll leave an unused api enabled ...

csieber (approved):
nice cleanup!

PR #797, opened 2026-07-09
✨ token-vendor: Allow any HTTP method on verify endpoints behind feature flag

Tobias-Pe:
When `allow_any_method=true` (`default: true` via `token_vendor.allow_any_method` in `values-cloud.yaml`), `verifyTokenHandler` (`/apis/core.token-vendor/v1/token.verify`) and `verifyJWTHandler` (`/apis/core.token-vendor/v1/jwt.verify`) permit any HTTP method (`GET`, `POST`, `PUT`, `DELETE`, `PATCH`).

#### Why

1. Under Gateway API (`use_istio=true`), Envoy`s external authorization (`ext_authz`) preserves the exact HTTP method of the downstream request when forwarding the token check to `token-vendor`.
2. When `token-vendor` restricted checks to `GET` only (`r.Method != http.MethodGet`), reverse relay streams (`POST /frontend/server/response`) and API requests were rejected by `token-vendor` with `400 Bad Request`.
3. The verification handlers (`VerifyToken` and `isValidJWT`) are purely read-only and idempotent. Enabling `--allow_any_method=true` by default (`token_vendor.allow_any_method: true` in `values-cloud.yaml`) allows `ext_authz` token verification checks across all HTTP methods while allowing operators to disable (`token_vendor.allow_any_method: false`) if strict `GET` checking is required.

- Add `--allow_any_method` boolean flag (`default: false`) to `src/go/cmd/token-vendor/main.go`.
- Update `verifyTokenHandler` and `verifyJWTHandler` in `src/go/cmd/token-vendor/api/v1/v1.go` to bypass method check when `allowAnyMethod` is true.
- Add `token_vendor.allow_any_method: true` default value to `src/app_charts/token-vendor/values-cloud.yaml`.
- Add `- --allow_any_method=true` when `.Values.token_vendor.allow_any_method` is set in `src/app_charts/token-vendor/cloud/token-vendor.yaml`.

go/traj/f521d819-4257-4a9b-b9f8-53bfa47865e8
TESTED=bazel test //src/go/cmd/token-vendor/... //src/app_charts/token-vendor/... and on live private cluster
RELNOTES=NONE
TAG=agy
CONV=83b810ff-8c02-4bfd-a194-1057ee54b528

Pull Request: https://github.com/googlecloudrobotics/core/pull/797

### DIFF
--- a/src/go/cmd/token-vendor/api/v1/v1.go
+++ b/src/go/cmd/token-vendor/api/v1/v1.go
@@ -43,11 +43,12 @@ const (
 )
 
 type HandlerContext struct {
-	tv *app.TokenVendor
+	tv             *app.TokenVendor
+	allowAnyMethod bool
 }
 
-func NewHandlerContext(tv *app.TokenVendor) *HandlerContext {
-	return &HandlerContext{tv}
+func NewHandlerContext(tv *app.TokenVendor, allowAnyMethod bool) *HandlerContext {
+	return &HandlerContext{tv, allowAnyMethod}
 }
 
 // getQueryParam extracts a query parameter from the request URL.
@@ -319,7 +320,7 @@ func (h *HandlerContext) tokenOAuth2Handler(w http.ResponseWriter, r *http.Reque
 // - Authorization: JWT that allows authorization
 // Response: only http status code
 func (h *HandlerContext) verifyJWTHandler(w http.ResponseWriter, r *http.Request) {
-	if r.Method != http.MethodGet {
+	if !h.allowAnyMethod && r.Method != http.MethodGet {
 		api.ErrResponse(r.Context(), w, http.StatusMethodNotAllowed,
 			fmt.Sprintf("method %s not allowed, only %s", r.Method, http.MethodGet))
 		return
@@ -363,7 +364,7 @@ func (h *HandlerContext) verifyJWTHandler(w http.ResponseWriter, r *http.Request
 // Headers (optional): X-CRC-TV-ROBOTS, X_FORWARDED_ACCESS_TOKEN or AUTHORIZATION
 // See function `tokenFromRequest` for details on how to supply the token.
 func (h *HandlerContext) verifyTokenHandler(w http.ResponseWriter, r *http.Request) {
-	if r.Method != http.MethodGet {
+	if !h.allowAnyMethod && r.Method != http.MethodGet {
 		api.ErrResponse(r.Context(), w, http.StatusBadRequest,
 			fmt.Sprintf("method %s not allowed, only %s", r.Method, http.MethodGet))
 		return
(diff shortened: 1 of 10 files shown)


Tobias-Pe:

csieber:
Can we make this more narrow? We could just allow GET + the original method and add a TODO to remove the GET (if it is not the intended method)

Tobias-Pe:
@csieber Under Istio / Gateway API (crc-gateway), Envoy's ext_authz filter (crc-auth-gateway) intercepts every single request on the Gateway and forwards a subrequest to token-vendor (/apis/core.token-vendor/v1/token.verify). Crucially, Envoy ext_authz preserves the exact HTTP method of the downstream client request on the verification subrequest.

If a client sends POST /frontend/server/response, ext_authz sends POST /token.verify.
If a robot (cr-syncer) sends PUT /apis/core.kubernetes/api/v1/namespaces/default/status, ext_authz sends PUT /token.verify.
If a user sends DELETE /apis/apps/v1/..., ext_authz sends DELETE /token.verify.
Because verifyTokenHandler and verifyJWTHandler are purely read-only, idempotent verification endpoints (they read headers/tokens, verify signatures/access, and return Status 200 OK or error), allowing any method when allow_any_method=true (v1.go) ensures all ext_authz checks (PUT, DELETE, PATCH, POST, GET) succeed across all Gateway routes without risk. 

How would one go at finding out the original method? The token vendor doesn't know about the pre auth policy

csieber:
> How would one go at finding out the original method? The token vendor doesn't know about the pre auth policy

The token-vendor is specifying the contract, so it should be considered the source of truth that it accepts. The question is if there are clients which send the wrong method in the first place.

csieber:
Now I understand. You are right, it can not know. I take back my objection.

src/go/cmd/token-vendor/api/v1/v1.go
```diff
@@ -487,11 +488,11 @@ func isValidJWT(jwt string) (bool, error) {
 }
 
 // Register the API V1 API handler functions to the default http.DefaultServeMux
-func Register(tv *app.TokenVendor, prefix string) error {
+func Register(tv *app.TokenVendor, prefix string, allowAnyMethod bool) error {
 
 	slog.Debug("mounting API V1", slog.String("Prefix", prefix))
 
-	h := NewHandlerContext(tv)
+	h := NewHandlerContext(tv, allowAnyMethod)
```
therjak: i think you can remove most of your change if you use the full pattern for 'HandleFunc'. ([METHOD ][HOST]/[PATH]). so you would remove the Method check in your handlefunc and just add 'GET ' in front ouf your pattern if !allowAnyMethod.

csieber (approved):

src/go/cmd/token-vendor/api/v1/v1.go
Tobias-Pe: Done, Thanks good tip!

PR #801, opened 2026-07-10
Create Helm chart for chart-assignment-controller

drigz:
This packages the controller as a standalone Helm chart using rules_helm,
allowing it to be installed on vanilla Kubernetes clusters. We use rules_helm
(instead of the existing packaging method) because it supports OCI registries,
which is required to release the chart to Google Artifact Registry. Webhook and
cert-manager integration are enabled by default but can be disabled via a
single value if cert-manager is not available.

Tested as described in the README. I only verified startup, I didn't yet
try creating a ChartAssignment.


drigz:
This PR is part of a 3 changes series
* Create Helm chart for chart-assignment-controller
* https://github.com/googlecloudrobotics/core/pull/802
* https://github.com/googlecloudrobotics/core/pull/803

src/app_charts/chart-assignment-controller/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "chart-assignment-controller.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "chart-assignment-controller.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
ensonic: out of curiosity, is this handwritten or generated. Do we need all this indirection?

src/app_charts/chart-assignment-controller/templates/chartassignment-crd.yaml
```diff
@@ -0,0 +1,82 @@
+{{- if .Values.installCRDs -}}
```
ensonic, line 1: are we duplicating this? We must have this somewhere else already?

ensonic (commented):
I guess you intentionally copy the whole set of the resources, right? Any way we can avoid this? Not that we change this often, but it is not discoverable right now.

ensonic:
> I guess you intentionally copy the whole set of the resources, right? Any way we can avoid this? Not that we change this often, but it is not discoverable right now.

Nevermind, just saw that you dedup in the next PR.

ensonic (approved):

src/app_charts/chart-assignment-controller/templates/_helpers.tpl
drigz: The PR is generated, but this boilerplate comes from Helm upstream: https://github.com/helm/helm/blob/e007c900ce5f2233c8583565d5ba1b0433b1a213/pkg/chartutil/create.go#L427

I first kept it because I've seen it in so many other Helm charts, but on reflection it's unnecessary because we can have at most one chart-assignment-controller in a cluster. If we do need it in future it'd be better to add it in a separate PR and minimize the delta in this one.

PR #804, opened 2026-07-13
Restore default grafana password.

ensonic:
See https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/UPGRADE.md#from-78x-to-79x


drigz (approved):
Thanks! Added a comment with your note from the thread.

PR #805, opened 2026-07-13
Set cluster upgrade timeout based on node count

koonpeng:
Sometimes when upgrading a large number of nodes it may exceed the default timeout of 1h.


ensonic (approved):
Thanks!

PR #807, opened 2026-07-13
Move the validator code in approllout to own file

ensonic:
This follows the example from the chart assignment controller.

Also contains gazelle cleanup on previous changes.


koonpeng (approved):
lgtm

PR #813, opened 2026-07-14
Eliminate klog v1 dependency.

ensonic:
* Switch last use to klog/v2.
* Switch test to use slong, rather than klog.
* Add a comment why we added klog in cr-syncer/main.go

### DIFF
--- a/src/go/tests/k8s_integration_test_auth_helper.go
+++ b/src/go/tests/k8s_integration_test_auth_helper.go
@@ -35,18 +35,21 @@ import (
 	"sync"
 	"time"
 
+	"log/slog"
+	"os"
+
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/yaml"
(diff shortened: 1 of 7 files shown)


src/go/tests/k8s_integration_test_auth_helper.go
```diff
@@ -343,7 +345,7 @@ func (c *commandTokenSource) parseTokenCmdOutput(output []byte) (*oauth2.Token,
 	}
 	var expiry time.Time
 	if t, err := time.Parse(c.timeFmt, expiryStr); err != nil {
-		klog.V(4).Infof("Failed to parse token expiry from %s (fmt=%s): %v", expiryStr, c.timeFmt, err)
+		slog.Debug("Failed to parse token expiry", "expiry", expiryStr, "format", c.timeFmt, "error", err)
```
therjak: the use of slog.String("expiry", expiryStr), slog.String("format", c.timeFmt), slog.Any("error", err) makes the arguments easier to parse.

ensonic: Done for the whole file.

csieber (approved):

PR #814, opened 2026-07-14
Some minor optimization to NGINX shield to handle connection bursts

awhdesmond:
Some minor optimization to NGINX shield to handle connection bursts
- increase worker_connections
- increase upstream keepalive
- increase listen backlog


ensonic (approved):
Thanks

PR #818, opened 2026-07-17
Add readiness probes to Deployments lacking them.

ensonic:
Insert readiness probes for Deployments that have liveness probes but no readiness probes, targeting the same endpoint. This ensures that pods do not receive traffic until they are actually ready, improving update safety. The probes use a periodSeconds of 5 and failureThreshold of 2 to balance quick readiness detection with reasonable probing frequency.

### DIFF
--- a/src/app_charts/token-vendor/cloud/token-vendor.yaml
+++ b/src/app_charts/token-vendor/cloud/token-vendor.yaml
@@ -47,6 +47,12 @@ spec:
             path: /healthz
             port: 9090
           initialDelaySeconds: 15
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 9090
+          periodSeconds: 5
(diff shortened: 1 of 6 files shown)


src/app_charts/token-vendor/cloud/token-vendor.yaml
```diff
@@ -47,6 +47,12 @@ spec:
             path: /healthz
             port: 9090
           initialDelaySeconds: 15
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
```
csieber: Are you sure `8080` is correct here? The health endpoint has `9090`.

ensonic: fixed. thanks for spotting.

csieber (approved):

PR #819, opened 2026-07-20
default disk type to null; add deploy.sh plan

koonpeng:
- Defaults disk type to `null` to avoid changing existing clusters that uses `pd-standard`.
- Add `deploy.sh plan` to see the terraform changes before applying.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -39,6 +39,7 @@ fi
 
 TERRAFORM_DIR="${DIR}/src/bootstrap/cloud/terraform"
 TERRAFORM_APPLY_FLAGS=${TERRAFORM_APPLY_FLAGS:- -auto-approve}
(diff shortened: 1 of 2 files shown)


drigz (approved):
> Defaults disk type to null to avoid changing existing clusters that uses pd-standard.

It's a bit too late for this, right? Or are you hoping to fast-track this to QA today so it goes out with 20260720?

koonpeng:
Intentional is to avoid similar issues in the future.

ensonic:
Maybe worth a comment to explain why we reverted. "pd-balanced" is the default now and we only have 4 systems left.

We'd need to talk to QA and make sure we have this version on intrinsic-staging, so that the previous version is not going to be applied to any prod project.

ensonic:
@koonpeng all projects now use "pd-balanced"

PR #820, opened 2026-07-20
set_config add option to configure disk type

koonpeng:
If `default` or not set, `node_disk_type` var will be set to `null`, which will use `hyperdisk-balanced` for 4th gen nodes (n4, c4 etc) and `null` (fallback to GCP terraform module default) for other nodes.


ensonic:
I'd prefer to not add another options.

koonpeng:
> I'd prefer to not add another options.

Agreed, it should be very rare that we will need to customize the disk type.

PR #831, opened 2026-07-27
♻️ app_charts: migrate companion routes and remove crc-auth-gateway

Tobias-Pe:
Decommission crc-auth-gateway companion routes and update HTTPRoute manifests to use Istio AuthorizationPolicy resources.

#### Why

Replace legacy double-gateway rewriting rules with direct Istio AuthorizationPolicy targetRef targeting Gateway API routes.

#### The code changes include:

- Remove crc-auth-gateway Gateway and catchall deployment/service from istio.yaml
- Remove legacy -auth companion routes across base, k8s-relay, prometheus, and token-vendor
- Add AuthorizationPolicy resources targetRef'ing HTTPRoutes for external authentication providers

PR #837, opened 2026-07-28
✨ cr-syncer-auth-webhook: implement inline reverse proxy for kubernetes-api route

Tobias-Pe:
Implement an inline reverse proxy handler in cr-syncer-auth-webhook for /apis/core.kubernetes/ and update HTTPRoute/kubernetes-api to target cr-syncer-auth-webhook directly.

#### Why

Prevents Authorization header collisions (Bearer JWT + Bearer K8s Token) caused by default Envoy ExtAuthz header appending when proxying requests to the Kubernetes API server.

<details><summary>Details</summary>
<p>

1. **Header Replacement Cleanliness**: Standard Envoy `ext_authz` appends allowed headers using `APPEND_IF_EXISTS_OR_ADD`. When replacing a client's robot JWT with a Kubernetes ServiceAccount token (`<k8s-sa-token>`), Envoy creates duplicate header entries (`Authorization: Bearer <robot-jwt>, Bearer <k8s-sa-token>`), which causes Kubernetes API server authentication to fail. The Go reverse proxy cleanly strips the client `Authorization` header and replaces it with the internal ServiceAccount token.
2. **Eliminating Provider Conflicts**: Using Envoy `ext_authz` for K8s API token sync required declaring a separate `auth-provider-cr-syncer` extension provider on `crc-gateway`, triggering Istio's `provider_conflicts_same_selector` error when combined with other Gateway-level providers.
3. **Transparent Proxying**: `cr-syncer-auth-webhook` validates the client's identity/JWT, swaps the credentials, and streams requests and responses directly to `https://kubernetes.default.svc:443` without Envoy filter overhead.

During derisking and testing on live clusters, three major architectural constraints were identified when using Envoy `ext_authz` and Istio `AuthorizationPolicy` for specialized authentication flows (specifically Kubernetes Control-Plane ServiceAccount Token synchronization via `cr-syncer-auth-webhook`).

### 6.1 Gotcha 1: Default Envoy ExtAuthz `APPEND_IF_EXISTS_OR_ADD` Header Behavior

- **The Problem**: When Envoy's `ext_authz` HTTP filter receives an HTTP `200 OK` from an external authorization service along with response headers listed under `headersToUpstreamOnAllow` (such as `authorization`), Envoy's default header handling strategy is `APPEND_IF_EXISTS_OR_ADD`.
- **Impact on Bearer Tokens**: If the downstream client sends an `Authorization` header containing a robot JWT (`Bearer <robot-jwt>`), and the external auth provider attempts to replace it with a Kubernetes ServiceAccount token (`Bearer <k8s-sa-token>`), Envoy does **not** replace the header. Instead, it appends the new value, producing a malformed comma-separated header:
  ```http
  Authorization: Bearer <robot-jwt>, Bearer <k8s-sa-token>
  ```
- **Failure Mode**: When passed to upstream backends (such as the Kubernetes API server), the backend fails to parse the duplicated `Authorization` header and rejects requests with HTTP `401 Unauthorized` or HTTP `400 Bad Request`.

### 6.2 Gotcha 2: Istio Listener Instantiation Constraint

- **The Problem**: Istio Pilot constructs Envoy HTTP filter chains lazily based on declared `AuthorizationPolicy` manifests.
- **Requirement for Baseline Policy**: Istio **only** instantiates the `envoy.filters.http.ext_authz` filter into an Envoy Gateway's active listener pipeline if there is at least one `AuthorizationPolicy` with `action: CUSTOM` attached directly to the Gateway resource (`Gateway/crc-gateway` in namespace `default`).
- **Failure Mode**: If service teams attempt to attach route-specific `action: CUSTOM` policies via `targetRefs` without a baseline `CUSTOM` policy applied at the `Gateway` level, Istio fails to compile `ext_authz` into Envoy's HTTP connection manager filter chain. As a result, authorization checks are entirely skipped, or routing fails silently.

### 6.3 Gotcha 3: Istio 1-Provider-Per-Workload Limit (`provider_conflicts_same_selector`)

- **The Problem**: Istio enforces a hard restriction where a single workload (e.g. `crc-gateway` pod selector) can only be bound to **one** `CUSTOM` action `extensionProvider` at the workload/Gateway level.
- **Provider Conflict**: If multiple Gateway-level `AuthorizationPolicy` manifests attempt to bind different providers (e.g., `auth-provider-default` for general API traffic and `auth-provider-cr-syncer` for control plane sync) to `Gateway/crc-gateway`, Istio Pilot logs a critical configuration conflict:
  ```text
  [Istio Pilot Error] provider_conflicts_same_selector: multiple custom extension providers targeted at workload crc-gateway (auth-provider-default vs auth-provider-cr-syncer)
  ```
- **Failure Mode**: Istio Pilot rejects or ignores the conflicting policy, leaving the gateway with unpredictable or missing authorization rules.

### 6.4 Why Go Reverse Proxy in `cr-syncer-auth-webhook` is the Recommended Solution

To cleanly eliminate all three gotchas, the recommended architecture decouples Kubernetes API token replacement from Envoy's `ext_authz` pipeline by implementing `cr-syncer-auth-webhook` as an inline **Go Reverse Proxy** (`httputil.ReverseProxy`):

| Gotcha / Issue | Envoy ExtAuthz Approach | Go Reverse Proxy Solution in `cr-syncer-auth-webhook` |
| :--- | :--- | :--- |
| **Header Mutation Collision** | Appends headers (`Bearer <token1>, Bearer <token2>`) | **Clean Replacement**: Go code explicitly deletes `Authorization` header and sets `Bearer <k8s-sa-token>`. |
| **Listener Instantiation** | Requires baseline `CUSTOM` policy on `Gateway/crc-gateway` | **Direct Routing**: Route `HTTPRoute/kubernetes-api` points directly to `cr-syncer-auth-webhook` as backend. |
| **Provider Conflicts** | Causes `provider_conflicts_same_selector` on `crc-gateway` | **Zero ExtAuthz Overhead**: No need to register `auth-provider-cr-syncer` as a competing Gateway provider. |

By routing `/apis/core.kubernetes.io` via `HTTPRoute` directly to `cr-syncer-auth-webhook`, `cr-syncer-auth-webhook` authenticates incoming client JWTs, injects internal K8s SA tokens, and reverse proxies the traffic to `https://kubernetes.default.svc:443` cleanly and efficiently.

</p>
</details> 

- Add /apis/core.kubernetes/ reverse proxy handler in main.go and validateRequestPath in request.go
- Add unit tests in main_test.go, request_test.go, and ext_authz_test.go
- Update HTTPRoute/kubernetes-api in kubernetes-api.yaml to route directly to cr-syncer-auth-webhook:80

go/traj/0fd13416-78f3-446a-8c5a-5272816fc88e
TESTED=UNITTESTS
RELNOTES=NONE
TAG=agy
CONV=0fd13416-78f3-446a-8c5a-5272816fc88e


Tobias-Pe:
This PR is part of a 2 changes series
* ✨ cr-syncer-auth-webhook: implement inline reverse proxy for kubernetes-api route
* https://github.com/googlecloudrobotics/core/pull/831

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -106,14 +141,13 @@ func (h *handlers) resourceIsFiltered(groupKind string) bool {
...
 // only accesses allowed resources.
-func (h *handlers) validateRequest(r *http.Request, robotName string) error {
-	urlString := extractOriginalURL(r)
+func (h *handlers) validateRequestPath(urlString string, robotName string) error {
 	incomingReq, err := parseURL(urlString)
 	if err != nil {
-		slog.Error("unexpected value of origin URL header",
-			slog.String("URL", urlString), ilog.Err(err))
+		slog.Error("unexpected value of target URL path",
+			slog.String("url", urlString), ilog.Err(err))
```
ensonic: any reason why changing them to start lower case? If there is a convention, we should do a post cleanup on the whole repo.

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -130,16 +164,82 @@ func (h *handlers) validateRequest(r *http.Request, robotName string) error {
...
+		http.Error(w, "Request not allowed", http.StatusForbidden)
+		return
+	}
+
+	k8sToken, err := os.ReadFile(*k8sTokenPath)
+	if err != nil {
+		slog.Error("failed to read serviceaccount token", ilog.Err(err))
+		http.Error(w, "Internal error", http.StatusInternalServerError)
+		return
+	}
```
ensonic: can we read this once (e.g. in a closure wrapping the handler func?

ensonic: That could also help to avoid that we have to make it a flag. I think having this being a package-level global var would in any case be enough

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -140,6 +182,75 @@ func (h *handlers) validateRequest(r *http.Request, robotName string) error {
...
+func (h *handlers) extractRobotName(encodedJWT string) (string, error) {
+	if strings.HasPrefix(encodedJWT, legacyTokenPrefix) {
+		return "", nil
+	}
+	claims, err := jws.Decode(encodedJWT)
+	if err != nil {
+		slog.Error("failed to parse JWT despite previous verification")
+		return "", err
+	}
+	slog.Debug("JWT parsed", slog.String("ID", claims.Sub))
```
Ongy: Do you want to keep the debug statement?

If so, I think it's worth passing in the context and using the `slog.DebugContext` function so we can associate it with traces (now or later, not sure how much otel we got in CRC)

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -56,13 +69,43 @@ const (
...
+		client:       &http.Client{},
+		k8sTransport: transport,
+		k8sTargetURL: targetURL,
+	}
 }
 
-func newHandlers() handlers {
-	return handlers{
-		client: &http.Client{},
+func (h *handlers) getK8sToken() (string, error) {
```
Ongy: This is the `AutomountServiceAccountToken` way to get the token, right?

While we are here, could we move to a more cloud-native token source?

Ongy: If it's a lot of effort let's not. But if it's easy, it might be worth it.

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -140,6 +182,75 @@ func (h *handlers) validateRequest(r *http.Request, robotName string) error {
...
+			req.URL.Scheme = h.k8sTargetURL.Scheme
+			req.URL.Host = h.k8sTargetURL.Host
+			req.URL.Path = strings.TrimPrefix(req.URL.Path, "/apis/core.kubernetes")
+			if req.URL.Path == "" {
+				req.URL.Path = "/"
+			}
+			// Why: RawPath stores the original encoded path. Resetting RawPath = "" forces
+			// httputil.ReverseProxy to re-encode req.URL.Path without retaining the stripped prefix.
+			req.URL.RawPath = ""
+			req.Header.Set("Authorization", "Bearer "+k8sToken)
```
Ongy: This seems to be the only request specific bit of this proxy and the only reason we keep the k8sTransport on the struct.

Is there no easier way to replace a header on the request before we pass it on that allows us to keep the httputil.ReverseProxy around?

It feels wrong to recreate it every time.

Ongy (changes requested):
Minor mostly style comments.

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -158,30 +269,34 @@ func (h *handlers) auth(w http.ResponseWriter, r *http.Request) {
...
-	k8sToken, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
+	k8sToken, err := h.getK8sToken()
 	if err != nil {
-		slog.Error("failed to read /var/run/secrets/kubernetes.io/serviceaccount/token", ilog.Err(err))
+		slog.Error("failed to read serviceaccount token", ilog.Err(err))
 		http.Error(w, "Internal error", http.StatusInternalServerError)
 		return
 	}
-	w.Header().Add("Authorization", "Bearer "+string(k8sToken))
+	w.Header().Add("Authorization", "Bearer "+k8sToken)
```
awhdesmond: nit: make "Bearer " a const

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Done! Updated extractRobotName to take ctx context.Context and switched to slog.DebugContext and slog.ErrorContext.

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Keeping the mounted SA token file for now — kubelet automatically rotates projected token files on disk, so reading it per request works reliably for the internal K8s API server without adding client-go dependencies or extra RBAC TokenRequest permissions.

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Done! Constructed httputil.ReverseProxy once during initialization in newHandlers() (h.k8sProxy) so we avoid creating a new proxy object on every request.

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Done! Added const bearerPrefix = "Bearer " and replaced all occurrences across the package.

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -182,16 +206,16 @@ func (h *handlers) validateRequestPath(urlString string, robotName string) error
...
 
-func (h *handlers) extractRobotName(encodedJWT string) (string, error) {
+func (h *handlers) extractRobotName(ctx context.Context, encodedJWT string) (string, error) {
 	if strings.HasPrefix(encodedJWT, legacyTokenPrefix) {
 		return "", nil
 	}
 	claims, err := jws.Decode(encodedJWT)
 	if err != nil {
-		slog.Error("failed to parse JWT despite previous verification")
+		slog.ErrorContext(ctx, "Failed to parse JWT despite previous verification")
```
Ongy: We should log the actual error here. It's an unexpected case so we'll want to know what's going wrong.

Ongy (approved):

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Done! Included ilog.Err(err) in slog.ErrorContext.

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -47,22 +51,85 @@ var (
...
 }
 
-func newHandlers() handlers {
-	return handlers{
+func newHandlers() *handlers {
+	// Why: Cloning http.DefaultTransport preserves tuned connection pool defaults
+	// (TCP keep-alives, ALPN negotiation, MaxIdleConnsPerHost) while allowing us
+	// to attach custom TLS Root CAs for the internal K8s API server target.
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	if caCert, err := os.ReadFile(*k8sCAPath); err == nil {
```
csieber: signal boost that this is NO error check

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -47,22 +51,85 @@ var (
...
+		certPool.AppendCertsFromPEM(caCert)
+		transport.TLSClientConfig = &tls.Config{
+			RootCAs: certPool,
+		}
+	}
+
+	targetURL, err := url.Parse(*k8sTarget)
+	if err != nil {
+		slog.Error("Failed to parse k8s target URL", ilog.Err(err))
+		os.Exit(1)
```
csieber: Can we rather return an error instead of exciting mid-function here?

Tobias-Pe: Whats the expected behavior after returning an error ? 
The endpoint is not able to funtion with the wrong k8sTarget

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -47,22 +51,85 @@ var (
...
+	}
+
+	h := &handlers{
 		client: &http.Client{},
 	}
+
+	h.k8sProxy = &httputil.ReverseProxy{
+		Director: func(req *http.Request) {
```
csieber: https://go.dev/src/net/http/httputil/reverseproxy.go says "Director is deprecated. Use Rewrite instead."

src/go/cmd/cr-syncer-auth-webhook/main.go
```diff
@@ -47,22 +51,85 @@ var (
...
+			req.Host = targetURL.Host
+		},
+		Transport:     transport,
+		FlushInterval: -1, // Why: Flush immediately to prevent buffering on long-lived K8s API Watch streams.
+	}
+
+	return h
+}
+
+func (h *handlers) getK8sToken() (string, error) {
```
csieber: `get` is discouraged, can be shortened

Tobias-Pe:
@drigz csieber told me you have a lot of knowledge about the cr syncer. Could i get your review on this too 🙏

src/go/cmd/cr-syncer-auth-webhook/main.go
drigz: Returning an error makes it easier for main() to use defer(). Alternatively you can call this newHandlersOrDie so the behavior is not hidden.

drigz (approved):
Thanks! For future reference: This PR would be easier to review if it didn't contain changes like the case of log messages or replacing fmt.Errorf with errors.New.

src/go/cmd/cr-syncer-auth-webhook/main.go
Tobias-Pe: Changed to return an error. SGTM

PR #838, opened 2026-07-28
Migrate OpenCensus to OpenTelemetry

amanakin:
Migrate telemetry from OpenCensus to OpenTelemetry with backward-compatible propagators

Details:

* Dual propagation:
  Introduced `telemetry.HTTPPropagator` and `telemetry.GRPCPropagator` in `src/go/pkg/telemetry/tracing.go`. 
  - `HTTPPropagator` injects and extracts both W3C `traceparent` (OTel) and B3 (`X-B3-TraceId`, OpenCensus HTTP) headers.
  - `GRPCPropagator` injects both W3C `traceparent` and `grpc-trace-bin` (`ocprop.Binary`).
  - Dual propagation guarantees uninterrupted trace context continuity across hybrid services running mixed OpenCensus and OpenTelemetry versions.

* OpenTelemetry resources:
  Configured `sdktrace.TracerProvider` with `resource.WithAttributes(semconv.ServiceName(serviceName))`. Attaching `service.name` at the provider level removes repetitive per-span attribute tagging.

* Fixed relay context propagation bug:
  In `http-relay-client` (`client.go`), ensured `req.WithContext(backendCtx)` is set prior to calling `local.Do`. Previously, the unattached context caused HTTP transport clients (`ochttp`) to fall back to `context.Background()`, generating an orphaned root span and overwriting intended downstream trace headers.

Testing
- Verified deployment and trace propagation across `http-relay-client` and `http-relay-server` workloads in GKE clusters and onprem devices.


amanakin:

src/go/cmd/cr-syncer/main.go
```diff
@@ -94,44 +102,24 @@ var (
...
+			sdkmetric.Stream{Aggregation: sdkmetric.AggregationExplicitBucketHistogram{Boundaries: sizeBoundaries}},
+		),
+		sdkmetric.NewView(
+			sdkmetric.Instrument{Name: httpconv.ClientResponseBodySize{}.Name()},
+			sdkmetric.Stream{Aggregation: sdkmetric.AggregationExplicitBucketHistogram{Boundaries: sizeBoundaries}},
+		),
+		sdkmetric.NewView(
+			sdkmetric.Instrument{Name: httpconv.ClientRequestDuration{}.Name()},
+			sdkmetric.Stream{Aggregation: sdkmetric.AggregationExplicitBucketHistogram{Boundaries: latencyBoundaries}},
+		),
```
ensonic, line 122: I guess we can't do much about these metrics changing, but since we not even have a dashboard for this is probably ok.

src/go/cmd/cr-syncer/main.go
```diff
@@ -285,15 +272,36 @@ func main() {
...
+	r, err := resource.New(
+		ctx,
+		resource.WithAttributes(semconv.ServiceName("cr-syncer")),
+		resource.WithFromEnv(),
+		resource.WithTelemetrySDK(),
+	)
+	if err != nil {
+		slog.Error("resource.New", ilog.Err(err))
+		os.Exit(1)
+	}
```
ensonic, line 290: Could this also use the new telemetry.SetupCloudtracing()?

ensonic (approved):
Thanks a lot. This looks more involved than I had expected.

src/go/cmd/cr-syncer/main.go
amanakin: Not really. It doesn't setup cloud exporter, only local hook for zpage traces

src/go/cmd/cr-syncer/main.go
amanakin: Yeah, it's unnecessarily complicated to keep the old names. This would literally be a copy-paste of code from OpenCensus to emit old metrics manually

PR #839, opened 2026-08-03
✨ bootstrap: forward optional flags in run-install.sh

koonpeng:
Refactor argument parsing in `run-install.sh` to support forwarding optional deploy flags (such as `-f` / `--values`) to `deploy.sh` and `set-config.sh`.

#### Why

Allow callers of `run-install.sh` to pass additional flags (e.g. Helm values files like `-f values-envoyExtAuthzHttp.yaml`) through to `deploy.sh create`, where they are passed to `helm template`.

- Use `shift` in `run-install.sh` to parse `<project_id>`, optional `<version>`, and optional `<command>` while retaining any trailing deploy flags in `"$@"`
- Forward `"$@"` to `deploy.sh` (`create`, `update_infra`, `delete`) and `set-config.sh`

### DIFF
--- a/src/bootstrap/cloud/run-install.sh
+++ b/src/bootstrap/cloud/run-install.sh
@@ -20,7 +20,9 @@ set -o pipefail -o errexit
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 BUCKET_URI=${BUCKET_URI:-"https://storage.googleapis.com/cloud-robotics-releases"}
-if [[ $# -lt 1 || -z "$1" || "$1" == -* ]]; then
+GCP_PROJECT_ID="$1"
+
+if [[ -z "${GCP_PROJECT_ID}" || "${GCP_PROJECT_ID}" == -* ]]; then
   echo "Usage: $0 <project id> [<version-file>|<tarball>] [<command>] [deploy-args...]"
   echo "Supported commands:"
   echo "  --set-config    Updates the cloud config interactively."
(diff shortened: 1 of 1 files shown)


src/bootstrap/cloud/run-install.sh
```diff
@@ -20,27 +20,32 @@ set -o pipefail -o errexit
...
   echo "  --set-config    Updates the cloud config interactively."
   echo "  --set-oauth     Enables and configures OAuth interactively."
   echo "  --delete        Deletes Cloud Robotics from the cloud project."
   echo "  --terraform     Apply only terraform changes."
-  echo "  (default)       Install the specifies version."
+  echo "  (default)       Install the specified version."
   exit 1
 fi
 
+GCP_PROJECT_ID="$1"
```
Tobias-Pe: Why move this codeblock downwards ?

koonpeng: Moved back to the original place, not sure why gemini likes to move it down.

Tobias-Pe:
Pls supply how you tested this change. Otherwise looks solid. Thanks for the fix, I overlooked this apparently :)

koonpeng:
> Pls supply how you tested this change. Otherwise looks solid. Thanks for the fix, I overlooked this apparently :)

It's tested for syntax errors with `bash -n`. Haven't actually tested in an actual run, but the bug is discovered when trying running some internal deployment scripts, will test again after PR is merged.

Tobias-Pe (approved):
Thanks!

LGTM GIF

awhdesmond (approved):

PR #840, opened 2026-08-03
Add clarifying comment to optional CA cert err == nil check

Tobias-Pe:
Created using jj-spr


Tobias-Pe:

PR #848, opened 2026-08-06
🥅 synk: treat NoMatchError as transient error

Tobias-Pe:
Treat missing CRD discovery and REST mapping errors (meta.IsNoMatchError) as transient in Synk to automatically retry when new CRDs are dynamically installed.

#### Why

When a ChartAssignment references resources whose CRD is not yet available on the cluster (e.g. Istio or Gateway API resources), Synk fails with `get REST mapping: no matches for kind`. Treating this as non-transient caused chart-assignment-controller to cache the failure without retrying, leaving ChartAssignments stuck in Failed phase even after the CRD was subsequently deployed by base-cloud.

- Add `meta.IsNoMatchError(err)` to `IsTransientErr` in `src/go/pkg/synk/synk.go`
- Add unit test cases for `NoKindMatchError` and `NoResourceMatchError` in `src/go/pkg/synk/synk_test.go`

go/traj/9725df8e-fd47-489e-b7c2-487e27e381c1
TESTED=UNITTESTS
RELNOTES=NONE
TAG=agy
CONV=9725df8e-fd47-489e-b7c2-487e27e381c1


Tobias-Pe:
@drigz does this change make sense to you ? The idea is to resync especially if crd's have been missing last time. For unblocking the release i did

```bash
kubectl --context gke_robco-integration-test_europe-west1-c_cloud-robotics delete pod -l app=chart-assignment-controller
```

So the sync gets retriggered from the last state, where e.g. tokenvendor was missing the authzpolicy crd. 

This change should make sync trigger automatically

drigz (approved):

drigz:
> does this change make sense to you ?

I approved as it seems like it technically matches the intent of the line above. I can imagine that at some point this could mask a non-transient missing CRD error, but if we encounter that in the real world, it might be nice to detect it after a few retries and surface it to the user with a dedicated error message.

PR #849, opened 2026-08-07
Accept prometheus external label object 

liyin00:
**Description**
This change allows users to configure externalLabels on the Prometheus Custom Resource for both cloud and robot deployments. These labels will be attached to all metrics when communicating with external systems (e.g., during remote write, federation, or when using Thanos), which is extremely useful for uniquely identifying the source of the metrics.

During chart rendering, these labels are dynamically parsed and injected into the underlying Prometheus instance. If no labels are provided, the configuration will gracefully default to an empty configuration {}.

**How to use**
Users can now define multiple external labels as a dictionary in their values file under prometheus.prometheusSpec.externalLabels.

Example usage in your custom values.yaml, values-cloud.yaml, or values-robot.yaml:

```
prometheus:
  prometheusSpec:
    externalLabels:
      project_id: my-gcp-project
      cluster: my-cluster-name
      environment: production
```


src/app_charts/prometheus/prometheus-cloud.values.yaml
```diff
@@ -60,6 +60,8 @@ prometheus:
         operator: "DoesNotExist"
 
     externalUrl: "${EXTERNAL_URL}"
+    externalLabels: 
+      project_id: "${GCP_PROJECT_ID}"
```
koonpeng: I think better to expose the entire `externalLabels`, iirc do something like `toYAML` (ai should know).

koonpeng (changes requested):

src/app_charts/prometheus/prometheus-cloud.values.yaml
liyin00: updated, thanks!

koonpeng (approved):
lgtm

ensonic (approved):
Having a bit more context in the cl description would be nice.

ensonic (approved):
Have a bit more context in the pr description would be nice.

src/app_charts/prometheus/cloud/prometheus-operator.yaml
```diff
@@ -29,4 +29,20 @@
...
+# Inject the externalLabels as a pre-formatted YAML block to allow users to define multiple labels in a dict
+{{- $prometheusExternalLabelsMap := dict -}}
+{{- if .Values.prometheus -}}
+  {{- if .Values.prometheus.prometheusSpec -}}
+    {{- $prometheusExternalLabelsMap = .Values.prometheus.prometheusSpec.externalLabels -}}
+  {{- end -}}
+{{- end -}}
+{{- if $prometheusExternalLabelsMap }}
+  {{- $prometheusExternalLabelsRawYaml := toYaml $prometheusExternalLabelsMap -}}
+  {{- $formattedExternalLabelsObject := $prometheusExternalLabelsRawYaml | nindent 6 -}}
```
ensonic: do you want to inline:
```
{{- $formattedExternalLabels := toYaml $prometheusExternalLabelsMap | nindent 6 -}}
```
one less line and shorter variables? Also below.

liyin00: updated this and the pr description on how to make use of this change. thank you!

PR #851, opened 2026-08-07
Drop more per-pod metrics

drigz:
These are the most expensive metrics that scale with the number of
pods, so I hope this will keep Prometheus alive on large setups until we
can migrate to VictoriaMetrics.

I'm not aware that these are used, and there's already a line above that
seems to try to drop them but is in the wrong place (which I'm not
removing to keep the change small as this file will soon be thrown
away).


drigz:

src/app_charts/prometheus/prometheus-robot.values.yaml
```diff
@@ -155,7 +154,7 @@ kubelet:
       regex: 'container_(file_descriptors|tasks_state|threads_max)'
     - sourceLabels: [__name__]
       action: drop
-      regex: 'container_spec.*|container_network.*'
+      regex: 'container_spec.*|container_network.*|container_blkio.*|container_fs.*'
```
ensonic, line 158: `container_fs.*` is used in
https://flowstate.intrinsic.ai/_p/giza-workcells/grafana/d/efa86fd1d0c121a26444b636a3f509a8/kubernetes-compute-resources-cluster?orgId=1&from=now-1h&to=now&timezone=utc&var-datasource=default&var-cluster=k10-h1-c04&refresh=10s

ensonic: Would it help if we sample less often?

drigz: Do you rely on those graphs? I guess we can live without them until we migrate to VictoriaMetrics.

We still get disk-level statistics from node-exporter and if needed can run `atop` in a privileged container to get granular disk stats (this relies on being able to reproduce issues though)

> Would it help if we sample less often?

I haven't tested but I don't believe so, much of the usage is driven by metric cardinality rather than sampling frequency.

What would help is dropping high-cardinality labels (the ones with UUIDs or IPs in as opposed to cluster/container name) but this is much harder to validate.

ensonic: I have been using these to rule out spikes in dis usage. We could squash the namespace here to reduce cardinality. Maybe with that we can also get some insights into network traffic back.

I have no data yet wheter victoria-metrics will handle this better (I hope it will).

ensonic (approved):
Approving to unblock, but I think we need to discuss alternatives.

src/app_charts/prometheus/prometheus-robot.values.yaml
drigz: I've made a note to check back on this once we have VictoriaMetrics onprem and see if we can re-enable.

PR #850, opened 2026-08-07
Make Grafana a standalone application

methylDragon:
Extract Grafana from the Prometheus application, enabling it to be deployed and managed independently.

!!! Merge AFTER ([context](https://chat.google.com/room/AAAA-ATNShc/WHZHvxyCf00)):
- https://github.com/intrinsic-ai/insrc/pull/51993

#### Why

So Grafana becomes a cross cutting concern, used by Prometheus and (soon to be) VictoriaMetrics

Previously, Grafana was deployed as a component of the Prometheus operator chart, which tightly coupled its lifecycle and configuration to Prometheus. This change decouples Grafana, allowing for:

-   **Independent Management:** Grafana can now be configured, deployed,
    and updated separately from Prometheus.
-   **Clearer Ownership:** Grafana resources are now owned by its own
    dedicated application definition.
-   **Reduced Conflicts:** Explicitly disables Grafana within the
    Prometheus chart and ensures CRD ownership is handled correctly,
    preventing resource conflicts.

#### The code changes include:

-   Adding a new `grafana` application definition under `src/app_charts`.
-   Moving Grafana's HTTPRoute and Ingress configurations to the new app.
-   Configuring the standalone Grafana to use the `kube-prometheus-stack`
    Helm chart, but with only Grafana components enabled.
-   Disabling Grafana within the `prometheus` application's chart.

PR #852, opened 2026-08-07
Add VictoriaMetrics + Grafana apps, Prometheus toggle/port config

Heet852003:
Migrating from Prometheus to VictoriaMetrics - two separate clusters, one for robot metrics and one for cloud metrics, since their growth patterns are different. Grafana's pulled out into its own standalone app too.

New apps: victoriametrics-robotmetrics (robot-metrics cluster - vminsert/vmselect/vmstorage/vmalert, plus the per-robot vmagent selector), victoriametrics-cloudmetrics (cloud-metrics cluster), grafana.
Enable/disable is controlled by AppRollout robots:/cloud: block presence, not explicit flags : simpler, matches how the CRD actually works.

Prometheus gets an enable toggle on both cloud and robot side (defaults true, no change for existing projects), plus a configurable remote-write port for when it needs to push into VictoriaMetrics during the hybrid period.

Also built out what Cloud Ops needs to pull our metrics into their Mimir cluster - an injectable sidecar for auth, a second remote_write target with correct positional relabel handling, external labels for their dedup keys (project_id, region), and a GCP service account annotation on the cloud vmagent's ServiceAccount. All of this is on both the cloud-metrics vmagent and the robot's own vmagent, since that's the only vmagent touching robot data at all.

Also moved kube-state-metrics and node-exporter from standalone vendored charts to kube-prometheus-stack, same reasoning as Grafana , avoids missing default config a standalone injection risks.Everything's default-off/inert unless explicitly set, so none of this changes behavior for any existing project.

Builds on top of @methylDragon's Grafana decoupling (core#850, insrc#51993).

RELNOTES=NONE


src/app_charts/victoriametrics-cloudmetrics/scripts/victoriametrics-cloudmetrics-approllout.yaml
methylDragon, line 1: Scripts should be removed and not checked in

src/app_charts/victoriametrics-cloudmetrics/kube-state-metrics.cloud.values.yaml
```diff
@@ -0,0 +1,12 @@
+# Configuration for the kube-state-metrics chart, added to satisfy
+# Brandon's confirmed requirement (July 20, 2026) that kube-state-metrics
```
methylDragon: ??? Why is this comment being checked in

src/app_charts/victoriametrics-cloudmetrics/victoriametrics-agent.cloud.values.yaml
```diff
@@ -0,0 +1,67 @@
+# Cloud-side vmagent, added July 20, 2026 per Brandon's confirmed
```
methylDragon: Leaking a lot of details in comments, can you audit the comments

src/app_charts/victoriametrics-cloudmetrics/values-cloud.yaml
```diff
@@ -0,0 +1,33 @@
+enable_cloud_cloud_victoriametrics: true
```
methylDragon: Can we verify that this is toggleable from the AppRollout?

src/app_charts/victoriametrics-cloudmetrics/values-cloud.yaml
```diff
@@ -0,0 +1,33 @@
...
+gcp_service_account: ""
+
+limits:
+  cpu: "2000m"
+  memory: "12Gi"
+requests:
+  storage: "100Gi"
+retention:
+  time: "90d"
+vminsert_replicas: 1
```
methylDragon, line 11: Similarly for these and the ports, did we verify if they're toggleable from the AppRollout? Is there a limitation if these are numbers instead of strings?

src/app_charts/victoriametrics-cloudmetrics/cloud/vmagent-operator.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.enable_cloud_cloud_victoriametrics }}
+# This includes all resources expanded from the victoria-metrics-agent
+# chart using the values in ../victoriametrics-agent.cloud.values.yaml.
+# This is the CLOUD-side vmagent instance (scrapes kube-state-metrics
+# and kubelet/cAdvisor running in the cloud cluster itself) - distinct
+# from the ROBOT-side vmagent (../robot/vmagent-operator.yaml), which
+# scrapes robot hardware exporters.
+{{- $data := .Files.Get "files/vmagent-chart.cloud.yaml" -}}
+{{- $data = $data | replace "${VMINSERT_PORT}" (toString .Values.vminsert_port) | replace "release-name-placeholder" .Values.release_name | replace "HELM-NAMESPACE" .Release.Namespace -}}
+{{- if .Values.upstream.sidecar_volume_mounts -}}
```
methylDragon, line 9: I think we're still defining individual sidecar value replaces.

Why can't we substitute the entire block using a yaml rule (seems prometheus-operator [does this](https://github.com/googlecloudrobotics/core/blob/main/src/app_charts/prometheus/cloud/prometheus-operator.yaml#L22-L45)), or string replace a generic variable that unpacks to any yaml in case we missed something?

src/app_charts/victoriametrics-robotmetrics/cloud/victoriametrics-operator.yaml
methylDragon, line 1: Let's call this victoriametrics-robotmetrics

And make sure to add a README making it very clear that cloudmetrics is for collecting metrics from cloud services (and only has componentes in the cloud), and robotmetrics is for collecting metrics from the robot (and has components that span the robot and the cloud).

src/app_charts/victoriametrics-cloudmetrics/cloud/vmagent-operator.yaml
methylDragon, line 1: Isn't this just vmagent.yaml?

src/app_charts/victoriametrics-robotmetrics/cloud/victoriametrics-operator.yaml
methylDragon, line 1: Also, missing vmagent-operator.yaml in this cloud directory? There should be a vmagent in the cloud as well for the robotmetrics right?

src/app_charts/victoriametrics-cloudmetrics/scripts/victoriametrics-cloudmetrics-approllout.yaml
Heet852003: removed, they were redundant anyway now that the real insrc AppRollouts exist

src/app_charts/victoriametrics-cloudmetrics/kube-state-metrics.cloud.values.yaml
Heet852003: my bad, cleaned it up

src/app_charts/victoriametrics-cloudmetrics/victoriametrics-agent.cloud.values.yaml
Heet852003: did a full pass, all clean now

src/app_charts/victoriametrics-cloudmetrics/values-cloud.yaml
Heet852003: moot now, pulled the flags entirely per your later comment

src/app_charts/victoriametrics-cloudmetrics/values-cloud.yaml
Heet852003: verified before pulling the flags, tested both a plain number and a quoted string override, both worked identically, no type issue

src/app_charts/victoriametrics-cloudmetrics/cloud/vmagent-operator.yaml
Heet852003: dug into it, prometheus-operator can do that because kube-prometheus-stack's chart already natively supports those fields. our vmagent chart (old 0.6.0) genuinely has zero native extension points, checked directly. converted everything to real structured yaml in the meantime though. once the version upgrade lands we can drop the anchor stuff entirely and go native like yours

src/app_charts/victoriametrics-robotmetrics/cloud/victoriametrics-operator.yaml
Heet852003: done, renamed + readme added on both repos. one thing to flag: only renamed the App itself, not the AppRollout , namespace comes from the AppRollout's own name, so keeping that stable means zero migration risk and the grafana urls we already verified stay valid. lmk if you want full symmetry instead

src/app_charts/victoriametrics-cloudmetrics/cloud/vmagent-operator.yaml
Heet852003: renamed to vmagent-operator.yaml, matches robot's existing naming. noticed you said "vmagent.yaml" without "-operator" though - kept the suffix to match robot/vmagent-operator.yaml exactly, say the word if you want it dropped from both

src/app_charts/victoriametrics-robotmetrics/cloud/victoriametrics-operator.yaml
Heet852003: reconfirmed cleanly, no vmagent anywhere under robotmetrics/cloud/, only the robot's own on-prem one touches robot data.

methylDragon: Let's have full symmetry

src/app_charts/victoriametrics-cloudmetrics/cloud/vmagent-operator.yaml
methylDragon: Sorry, yeah vmagent-operator*

src/app_charts/victoriametrics-robotmetrics/cloud/victoriametrics-operator.yaml
methylDragon: But don't we need the robot vmagent to send to a cloud robot-vmagent so we can route robot metrics to the victoriametrics-robotmetrics cluster instance (and also eventually to cloud mimir)?

Basically your robotmetrics/robot creates the vmagent on each robot, but what about the robot vmagent that's in the cloud?

image

src/app_charts/grafana/cloud/app.yaml
methylDragon, line 1: Base your PR on top of https://github.com/googlecloudrobotics/core/pull/850 and use its src/app_charts/grafana instead (replacing your src/app_charts/grafana

third_party/kube-state-metrics/kube-state-metrics-5.33.2.tgz
methylDragon, line 1: !!! For all subcharts in kube-prometheus-stack that you might use in victoriametrics (e.g. kube-state-metrics, node-exporter), I think there are default configurations we currently rely on that are part of kube-prometheus-stack that might warrant us using it instead of injecting the charts directly.

See my comment in: https://github.com/googlecloudrobotics/core/pull/850

I tried to do a standalone injection of grafana and it missed all those configs. I suspect it'll be the same with the other components that used to be part of kube-prometheus-stack that you extracted out.

---

Let's use kube-prometheus-stack, but just disable prometheus and any other components we're turning down, just like I did to get grafana working (e.g. here: https://github.com/googlecloudrobotics/core/pull/850/changes#diff-831c168d469137791822495d70dad85851c76f8433824fae989945f40ece7753R58-R69)

src/app_charts/grafana/cloud/app.yaml
Heet852003: done , dropped my own grafana app, using yours instead, squashed everything into one commit on top of your branch

third_party/kube-state-metrics/kube-state-metrics-5.33.2.tgz
Heet852003: done for both kube-state-metrics and node-exporter, on kube-prometheus-stack now with everything else disabled, same as your grafana pattern. found a  gap doing node-exporter, our old mirror was missing a cardinality drop prometheus actually applies, added that too. render-verified both

PR #859, opened 2026-08-12
🚧 envoy-gateway: transition from istio to envoy gateway

Tobias-Pe:
This PR migrates the infrastructure from Istio to Envoy Gateway to align with NGINX ingress requirements and fulfill mandatory ext_authz requirements.

#### Why

WIP
Derisking envoy gateway to fullfill extauthz requirements and align with the simplicity of nginx ingress.

- Replaced Istio CRDs with Envoy Gateway resources (`EnvoyProxy`, `GatewayClass`, `SecurityPolicy`).
- Implemented `ExtAuthzServer` in `token-vendor` to handle gRPC-based authorization.
- Added `ReferenceGrant` objects to allow cross-namespace authentication policy enforcement.
- Updated `deploy.sh` and Helm values to toggle `USE_GATEWAY` instead of `USE_ISTIO`.
- Configured `BackendTrafficPolicy` to disable the default 15s request timeout for better parity with NGINX.
- Updated `token-vendor` build files to include necessary gRPC and control-plane dependencies.
- Added extensive test coverage for the new Envoy-based gRPC authorization flow.

RELNOTES=NONE
SDK_CHANGES=No changes
TESTED=private cluster


src/app_charts/base/cloud/gateway.yaml
```diff
@@ -0,0 +1,101 @@
...
+  name: crc-gateway-proxy-config
+  namespace: envoy-gateway-system
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyService:
+        type: LoadBalancer
+        externalTrafficPolicy: Local
+        {{- if .Values.ingress_ip }}
```
ensonic: do we need the if .values ... - we must have this from terrafrom, otherwise nothing works

src/app_charts/k8s-relay/cloud/http-route.yaml
```diff
@@ -48,32 +49,70 @@ spec:
...
+      value: "false"
+    - name: scopes
+      value: "tokenvendor,tvhuman"
+    headersToExtAuth:
+    - authorization
+    - cookie
+    - x-forwarded-access-token
+    - x-forwarded-host
+    - apikey-token
+    - x-server-name
```
ensonic: should we put this list into the values.yaml and reference via .Value.authProxy.authHeaders
then we can control it in one place and have less repeated code in the AuthPolicies

src/go/cmd/token-vendor/api/v1/authz.go
```diff
@@ -0,0 +1,289 @@
...
+	if req == nil || req.GetAttributes() == nil {
+		return deniedResponse(typev3.StatusCode_Unauthorized, codes.Unauthenticated, "missing request attributes"), nil
+	}
+
+	isRobotReq := isRobot(req)
+
+	httpReq := req.GetAttributes().GetRequest().GetHttp()
+	var headers map[string]string
+	var pathStr string
+	if httpReq != nil {
```
ensonic:
```suggestion
	var headers map[string]string
	var pathStr string
	if httpReq := req.GetAttributes().GetRequest().GetHttp(); httpReq != nil {
```

src/go/cmd/token-vendor/api/v1/authz.go
```diff
@@ -0,0 +1,289 @@
...
+		if u, err := url.Parse(extPath); err == nil {
+			tokenStr = u.Query().Get("token")
+		}
+	}
+
+	if tokenStr == "" {
+		return deniedResponse(typev3.StatusCode_Unauthorized, codes.Unauthenticated, "missing authorization credentials"), nil
+	}
+
+	if endpoint == "token.verify" || strings.Contains(endpoint, "token.verify") || strings.Contains(extPath, "/token.verify") || strings.Contains(pathStr, "/token.verify") {
```
ensonic: should this maybe be string.HasPrefix instead of strings.Contains? Also below.

src/go/cmd/token-vendor/api/v1/authz.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	if isRobot {
+		if ok, _ := isValidJWT(tokenStr); ok {
+			return s.verifyJWT(ctx, tokenStr)
+		}
+		if ok, _ := isValidToken(tokenStr); ok {
+			return s.verifyOAuthToken(ctx, tokenStr, true)
+		}
+		return deniedResponse(typev3.StatusCode_BadRequest, codes.InvalidArgument, "invalid token format"), nil
```
ensonic, line 168: I am not 100% sure here. When we passed robots=true to query string we still allowed humans.

src/go/cmd/token-vendor/api/v1/authz.go
```diff
@@ -0,0 +1,289 @@
...
+		if val, ok := ext["type"]; ok {
+			v := strings.ToLower(strings.TrimSpace(val))
+			if v == "robot" || v == "robots" {
+				return true
+			}
+			if v == "human" || v == "user" {
+				return false
+			}
+		}
+		if scopes, ok := ext["scopes"]; ok {
```
ensonic: do we have scopes in crc? If this is a comma separated value, I'd first split and look for a whole match

src/go/cmd/token-vendor/main.go
```diff
@@ -96,6 +101,10 @@ var (
 		"If true, token verification endpoints (token.verify and jwt.verify) allow any HTTP method (GET, POST, etc.) for compatibility with Gateway API ext_authz.")
 )
 
+func init() {
+	flag.IntVar(grpcPort, "grpc-port", 9091, "Port number for gRPC ext_authz server (alias for grpc_port)")
```
ensonic: move to main()

third_party/envoy-gateway/README.md
```diff
@@ -0,0 +1,121 @@
+# gateway-helm
```
ensonic: did you simply copy the upstream docs?

deploy.sh
```diff
@@ -88,7 +88,7 @@ function prepare_source_install {
       //src/go/cmd/setup-robot:setup-robot.push
 
   mkdir -p ${DIR}/bazel-bin/src/go/cmd/synk/synk_/
-  mv -n ${tmpdir}/synk ${DIR}/bazel-bin/src/go/cmd/synk/synk_/synk
+  mv -f ${tmpdir}/synk ${DIR}/bazel-bin/src/go/cmd/synk/synk_/synk
```
ensonic: why the change? "-n" == no clobber would only mv if it is not there already.

We can also do "mv -f", but then we can drop the "rm" in next line. Maybe best done as  a separate change though.

Tobias-Pe: Just a draft for now. I'll clean it up and split it all out once ready

src/app_charts/base/cloud/gateway.yaml
Tobias-Pe: Good point. Removed the {{- if .Values.ingress_ip }} check so loadBalancerIP is set directly, matching nginx-ingress-controller.yaml.

src/app_charts/k8s-relay/cloud/http-route.yaml
Tobias-Pe: Done. Added authProxy.authHeaders to values-cloud.yaml and templated headersToExtAuth in both SecurityPolicies via {{- toYaml .Values.authProxy.authHeaders | nindent 4 }}.

src/go/cmd/token-vendor/api/v1/authz.go
Tobias-Pe: Applied, thanks!

src/go/cmd/token-vendor/api/v1/authz.go
Tobias-Pe: Done. Switched to strings.HasPrefix for endpoint and added a hasPathSuffix helper that parses the URL path for extPath and pathStr to avoid loose substring matching.

src/go/cmd/token-vendor/main.go
Tobias-Pe: Done. Removed init() and moved the flag registration into main() before flag.Parse().

deploy.sh
Tobias-Pe: Reverted back to mv -n to keep this PR focused.

src/go/cmd/token-vendor/api/v1/authz.go
Tobias-Pe: That was halluzinated

third_party/envoy-gateway/README.md
Tobias-Pe: Changed to a real readme

src/go/cmd/token-vendor/api/v1/authz.go
Tobias-Pe: Jetski says 

### Legacy Behavior (`v1.go` HTTP Handler)

In [`src/go/cmd/token-vendor/api/v1/v1.go`](https://github.com/googlecloudrobotics/core/blob/main/src/go/cmd/token-vendor/api/v1/v1.go):

1. **Detection:**
   - The handler called `testForRobotACL(r.URL, &r.Header)` checking if `robots=true` was present in the query string or if `X-CRC-TV-ROBOTS: true` was set in the header.

2. **ACL Mapping:**
   - `robots=true` $\rightarrow$ Tested GCP IAM `iam.serviceAccounts.actAs` permission on `robot-service@<project>.iam.gserviceaccount.com`.
   - `robots=false` (or omitted) $\rightarrow$ Tested GCP IAM `iam.serviceAccounts.actAs` permission on `human-acl@<project>.iam.gserviceaccount.com`.

3. **Strict Rejection:**
   - If `robots=true` was set and IAM verification on `robot-service` failed (for example, if a human OAuth token was passed), it immediately returned `403 Forbidden` (`unable to verify token`). It never fell back to `human-acl`.

PR #863, opened 2026-08-14
✨ token-vendor: implement Envoy ext_authz v3 gRPC authorization server on :9091

Tobias-Pe:
Implement the Envoy envoy.service.auth.v3.Authorization gRPC service in token-vendor on port :9091.

#### Why

Unlike legacy NGINX Ingress which used HTTP forward-auth subrequests, Envoy Gateway performs external authorization natively via the envoy.service.auth.v3.Authorization gRPC protocol. token-vendor is CloudRobotics Core's central authentication authority for verifying API keys, robot-signed JWT assertions, and GCP OAuth tokens. Adding a dedicated gRPC ext_authz listener on port :9091 enables Envoy Gateway SecurityPolicy filters to validate incoming perimeter traffic synchronously without disrupting or modifying existing REST token exchange endpoints.

#### The code changes include:

- Implement ExtAuthzServer conforming to envoy.service.auth.v3.AuthorizationServer.
- Support validation for API key tokens, robot RSA JWT signatures, and GCP OAuth tokens.
- Add comprehensive unit tests in authz_test.go.
- Launch gRPC listener on port :9091 alongside HTTP server in main.go.
- Add Envoy control-plane and proto dependencies to go.mod, MODULE.bazel, and non_module_deps.bzl.

go/traj/467007a7-8537-4d71-b3f4-22b225909fb7
BUG: b/492106546
TESTED=private cluster
RELNOTES=NONE
TAG=agy
CONV=467007a7-8537-4d71-b3f4-22b225909fb7

PR #862, opened 2026-08-14
📦️ third_party: vendor Gateway API v1.2 and Envoy Gateway v1.8 generated CRD manifests

Tobias-Pe:
Vendor upstream Gateway API v1.2 and Envoy Gateway v1.8 CRDs and generated controller manifests.

#### Why

Envoy Gateway and Kubernetes Gateway API controllers cannot start or reconcile routing resources without their CustomResourceDefinitions being installed in the cluster. Vendoring these pre-generated upstream CRDs in an isolated PR separates 53,000+ lines of raw machine-generated YAML from functional code, allowing reviewers to verify the artifacts purely against the update scripts in PR 1.1 without bloating subsequent PR diffs.

#### The code changes include:

- Vendor standard Kubernetes Gateway API v1.2 CRDs (gatewayapi-crds.yaml).
- Vendor Envoy Gateway v1.8 CRDs (envoy-gateway-crds.yaml) and controller deployment manifest (envoy-gateway-generated.yaml).
- Remove obsolete grpcroute-generated.yaml.

go/traj/467007a7-8537-4d71-b3f4-22b225909fb7
BUG: b/492106546
TESTED=private cluster
RELNOTES=NONE
TAG=agy
CONV=467007a7-8537-4d71-b3f4-22b225909fb7

PR #864, opened 2026-08-14
🧱 base: deploy Envoy Gateway crc-gateway with tuned HTTP/2 and buffer settings

Tobias-Pe:
Deploy the Envoy Gateway controller, GatewayClass, crc-gateway, ClientTrafficPolicy, and BackendTrafficPolicy.

#### Why

With Istio being retired, the cluster requires a production Gateway API controller to provision the crc-gateway ingress data plane. Envoy Gateway's default parameters (32 KiB connection buffer limits, 15s request timeouts) severely bottleneck high-throughput streaming workloads and drop long-lived connections. Deploying the Envoy Gateway controller alongside tuned ClientTrafficPolicy (1 MiB connection buffer, 512 KiB HTTP/2 stream windows) and BackendTrafficPolicy (request timeouts disabled) establishes production performance and reliability before routing live traffic.

- Deploy Envoy Gateway controller via base cloud chart templates.
- Define GatewayClass and Gateway crc-gateway in gateway.yaml.
- Configure ClientTrafficPolicy with 1 MiB initial connection window and 512 KiB stream window.
- Configure BackendTrafficPolicy targeting crc-gateway with requestTimeout disabled.
- Add USE_GATEWAY feature flag in deploy.sh and values-cloud.yaml.
- Update cert-manager issuers and nginx-ingress manifests to support Gateway API cutover.

go/traj/467007a7-8537-4d71-b3f4-22b225909fb7
BUG: b/492106546
TESTED=private cluster
RELNOTES=NONE
TAG=agy
CONV=467007a7-8537-4d71-b3f4-22b225909fb7


Tobias-Pe:
This PR is part of a 4 changes series
* https://github.com/googlecloudrobotics/core/pull/863
* 🧱 base: deploy Envoy Gateway crc-gateway with tuned HTTP/2 and buffer settings
* https://github.com/googlecloudrobotics/core/pull/865
* https://github.com/googlecloudrobotics/core/pull/866

src/app_charts/base/cloud/gateway.yaml
```diff
@@ -0,0 +1,135 @@
...
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyService:
+        type: LoadBalancer
+        externalTrafficPolicy: Local
+        {{- if .Values.ingress_ip }}
+        loadBalancerIP: {{ .Values.ingress_ip }}
+        {{- end }}
```
ensonic: the ip must be set, we can skip the if

ensonic (approved):

Tobias-Pe:
@cloudops (@awhdesmond @liyin00 @koonpeng) PTAL at the envoy gateway configuration if the configuration is reasonable from an ops perspective 🙏

src/app_charts/base/cloud/gateway.yaml
Tobias-Pe: done :)

ensonic (approved):

src/app_charts/base/cloud/gateway.yaml
```diff
@@ -0,0 +1,134 @@
...
+                      app.kubernetes.io/name: envoy
+                  topologyKey: kubernetes.io/hostname
+        container:
+          resources:
+            requests:
+              cpu: 100m
+              memory: 256Mi
+            limits:
+              cpu: 2000m
+              memory: 1Gi
```
liyin00: i worry that this memory limit is too low for the bufferLimit of 5MiB, since that limit is on the r/w buffer for each connection. assuming a baseline Envoy overhead of ~200MiB, only ~824MiB remains for connection buffering. if connections reach their max buffer size, a single pod will OOMKill at just ~164 concurrent connections (824 / 5). shall we up the memory limits to match expected peak concurrency, or lower the buffer limits?

src/bootstrap/cloud/terraform/cluster.tf
```diff
@@ -44,7 +44,7 @@ resource "google_container_cluster" "cloud-robotics" {
   }
 
   gateway_api_config {
-    channel = "CHANNEL_STANDARD"
+    channel = "CHANNEL_DISABLED"
```
liyin00, line 47: will this affect the availability of any existing clusters? if so, i would recommend to shift this into the 4th part of your migration PRs to prevent any wipeouts of the routing configs

src/app_charts/base/cloud/gateway.yaml
```diff
@@ -0,0 +1,134 @@
...
+metadata:
+  name: crc-gateway-client-traffic-policy
+spec:
+  targetRefs:
+  - group: gateway.networking.k8s.io
+    kind: Gateway
+    name: crc-gateway
+  clientIPDetection:
+    xForwardedFor:
+      numTrustedHops: 1
```
liyin00: im a bit worried about setting this behind a L4 LB because the packets get passed without securely appending real client ip to header. this opens us up to ip spoofing - if an attacker sends request with xff `10.0.0.1`, envoy will trust that fake ip as source. 

if the goal here is to preserve IPs for service-to-service auth, i would recommend removing this to safely fall back to tcp socket add, and rely on safer mechanisms for establishing identity. wdyt?

src/bootstrap/cloud/terraform/cluster.tf
Tobias-Pe: in https://github.com/googlecloudrobotics/core/pull/862 we introduced the gateway api crds

src/app_charts/base/cloud/gateway.yaml
Tobias-Pe: Removed, let me test

src/app_charts/base/cloud/gateway.yaml
Tobias-Pe: Lowered it, Thanks !

awhdesmond (approved):
lgtm

src/app_charts/base/cloud/gateway.yaml
```diff
@@ -0,0 +1,131 @@
...
+  namespace: envoy-gateway-system
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyService:
+        type: LoadBalancer
+        externalTrafficPolicy: Local
+        loadBalancerIP: {{ .Values.ingress_ip }}
+      envoyDeployment:
```
awhdesmond, line 15: nit: maybe we can add a simple hpa?

Tobias-Pe: I tried to keep it close to nginx. Does nginx use a hpa ? If not let's make this a followup

awhdesmond: https://github.com/googlecloudrobotics/core/blob/main/src/app_charts/base/cloud/nginx-ingress-controller.yaml#L152

but yea, can be a followup

PR #868, opened 2026-08-17
Allow CRC config to be managed by terraform

koonpeng:
First step to support provisioning CRC via terraform module.

1. Add config to store whether configs are managed by terraform.
2. If managed by terraform, `set-config.sh` exits with error when trying to update config.

It shouldn't affect existing deployments without `CONFIG_MANAGED_BY_TERRAFORM`.


ensonic:
@koonpeng we can't eliminate the config.sh - too much of the tooling is built around it. I was wondering if we need a flag and if we manage the config via tf, we generate the config.sh file via terraform. Maybe we can make the bucket/file read-only.

src/bootstrap/cloud/terraform/gcs.tf
```diff
@@ -53,3 +53,43 @@ resource "google_storage_bucket_object" "config_store_crc_version" {
...
+  count = var.managed_config ? 1 : 0
+
+  byte_length = 16
+}
+
+resource "google_storage_bucket_object" "crc_config" {
+  count = var.managed_config ? 1 : 0
+
+  name          = "config.sh"
+  content       = <<EOF
```
koonpeng, line 67: This generates the config file from terraform. The deploy script sets `managed_config` to false. Direct provisioning should set it to true (default). Reading the config should be fine for current tooling, but updating the config will cause state drifts.

ensonic: Sorry could not access the changes yesterday due to the github downtime. So far I am mostly changing the configs through the crc-config tool (in insrc), we can make that aware of the variant (CONFIG_MANAGED_BY_TERRAFORM) and ensure that configs are changed via terraform.

scripts/config.sh
```diff
@@ -39,6 +39,11 @@ function save_variable {
   local name="$2"
   local value="$3"
 
+  if ( source "${config_file}" 2>/dev/null; [[ "${CONFIG_MANAGED_BY_TERRAFORM}" == "true" ]] ); then
```
ensonic:
```suggestion
  if ( source "${config_file}" 2>/dev/null; [[ "${CONFIG_MANAGED_BY_TERRAFORM:-}" == "true" ]] ); then
```
so that if a script uses "set -u" we set a default that is not "true".

ensonic (approved):

scripts/config.sh
koonpeng: done

PR #876, opened 2026-08-19
Fix Helm map coercion warnings for Prometheus charts

liyin00:
**Context**
This resolves the `coalesce.go:316: warning: cannot overwrite table with non table` warnings that were appearing in the presubmit. 
    
Previously, `externalLabels` and `nodeSelector` properties in our Helm values files were being overridden with string templates (e.g. `"${PROMETHEUS_EXTERNAL_LABELS}"`). As the upstream `kube-prometheus-stack` chart expects these to be maps, Helm produced warnings during coalescence. 
    
This change replaces the string template definitions in the values yaml files with dummy maps (e.g., `$PROMETHEUS_EXTERNAL_LABELS : REMOVE_ME`). The `replace` template rules in `prometheus-operator.yaml` have been updated to target these dummy keys to inject the processed configurations as before. This successfully aligns the data types to prevent Helm warnings from surfacing in `stderr` while keeping our template injection logic intact.

**Tests Run** 
This change has been helm-templated out with the following test values.yaml
```yaml 
prometheus:
  prometheusSpec:
    externalLabels:
      environment: test
      region: us-central1
```

and verified to render 
```yaml 
apiVersion: monitoring.coreos.com/v1
kind: Prometheus
metadata:
  name: kube-prometheus
  namespace: default
  labels:
    app: kube-prometheus
    
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: prom
    app.kubernetes.io/version: "87.5.1"
    app.kubernetes.io/part-of: kube
    chart: kube-prometheus-stack-87.5.1
    release: "prom"
    heritage: "Helm"
spec:
  automountServiceAccountToken: true
  image: "quay.io/prometheus/prometheus:v3.13.0-distroless"
  imagePullPolicy: "IfNotPresent"
  version: "v3.13.0-distroless"
  externalLabels:
    
      environment: test <-- successfully rendered
      region: us-central1 <-- successfully rendered
```

### DIFF
--- a/src/app_charts/prometheus/cloud/prometheus-operator.yaml
+++ b/src/app_charts/prometheus/cloud/prometheus-operator.yaml
@@ -14,9 +14,9 @@
 {{- if $prometheusNodeSelectorMap }}
   {{- $prometheusNodeSelectorRawYaml := toYaml $prometheusNodeSelectorMap -}}
   {{- $formattedPrometheusNodeSelectorObject := $prometheusNodeSelectorRawYaml | nindent 6 -}}
-  {{- $data = $data | replace "${CR_PROMETHEUS_NODE_SELECTOR_OBJECT}" $formattedPrometheusNodeSelectorObject -}}
+  {{- $data = $data | replace "$CR_PROMETHEUS_NODE_SELECTOR_OBJECT: REMOVE_ME" $formattedPrometheusNodeSelectorObject -}}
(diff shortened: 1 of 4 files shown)


ensonic (approved):
Thanks. I think we're hitting the limits with what we do here, but this fix is good for now.

PR #878, opened 2026-08-19
Fix race condition in http-relay-server unit tests

ensonic:
Initialize the backend state in the broker before starting the client request in TestClientHandler, TestClientHandlerWithChunkedResponse, TestRequestStreamHandler, and TestTrailers. Without pre-registering the backend channel, concurrent userClientRequest calls could fail immediately with 503 (unknown client), causing GetRequest to time out and panic on nil dereference.

Also change t.Errorf to t.Fatalf on GetRequest error checks to prevent nil pointer panics if request retrieval fails.

TAG=agy
CONV=d581302b-b390-4eef-bec8-72c58746e038


csieber (approved):
thanks!

PR #880, opened 2026-08-20
Prefix short SHA with 'g' in BUILD_VERSION for Helm chart

ensonic:
SemVer 2.0.0 Section 9 forbids leading zeros in numeric pre-release identifiers. When a commit short SHA consists only of digits and starts with a zero (such as 0156133), Helm fails chart metadata validation ('0.1.0-0156133').

Prefixing the short SHA with 'g' (e.g. 0.1.0-g0156133) follows the git describe convention and ensures the pre-release identifier is always alphanumeric, preventing invalid SemVer errors.

This is a followup for [pr/803](https://github.com/googlecloudrobotics/core/pull/880)

### DIFF
--- a/.github/ci/release_binary.sh
+++ b/.github/ci/release_binary.sh
@@ -43,7 +43,7 @@ release_binary "${GCP_BUCKET}" "crc-${VERSION}/crc-${VERSION}+${SHA}" ${LABELS}
 # - BUILD_VERSION is used by `bazel --stamp`.
 # - HELM_* are used by the push-chart script to auth to GAR.
 echo "Pushing Helm chart to GAR..."
-export BUILD_VERSION="${VERSION}-${SHA}"
+export BUILD_VERSION="${VERSION}-g${SHA}"
 export HELM_REGISTRY_USERNAME="oauth2accesstoken"
 export HELM_REGISTRY_PASSWORD=$(gcloud auth print-access-token)
 bazel_ci run //src/app_charts/chart-assignment-controller:push-chart


.github/ci/release_binary.sh
```diff
@@ -43,7 +43,7 @@ release_binary "${GCP_BUCKET}" "crc-${VERSION}/crc-${VERSION}+${SHA}" ${LABELS}
 # - BUILD_VERSION is used by `bazel --stamp`.
 # - HELM_* are used by the push-chart script to auth to GAR.
 echo "Pushing Helm chart to GAR..."
-export BUILD_VERSION="${VERSION}-${SHA}"
+export BUILD_VERSION="${VERSION}-g${SHA}"
```
Tobias-Pe, line 46: I'd go with:

```suggestion
export BUILD_VERSION="${VERSION}-crc-${SHA}"
```

The g prefix is just a bit too random imo

Tobias-Pe (approved):

.github/ci/release_binary.sh
ensonic: I adopted the standard git convention (git describe):
   * In the Git ecosystem, git describe automatically prefixes commit hashes with g (e.g., v0.1.0-4-g0156133), where g stands for "git".
   * It is the standard way tools differentiate a commit hash from version numbers or patch levels.

ensonic:
@drigz FYI. I don't know where you reference the helm chart on the IOC side, but you'll probably need to add the 'g'.

PR #881, opened 2026-08-20
👷 ci: allow 'Create release' workflow to pick up any passing postsubmit run on main

Tobias-Pe:
Allow 'Create release' workflow to pick up manual postsubmit runs when dispatched, rather than only scheduled cron runs.

#### Why

Previously, release.yml filtered postsubmit runs strictly on '&event=schedule'. This prevented manually dispatched postsubmits (created with force-binary-release) from being recognized when triggering Create release manually out-of-band.

- Add any-postsubmit boolean input to workflow_dispatch in .github/workflows/release.yml (defaulting to true).
- Conditionally remove the '&event=schedule' filter when triggered manually or when any-postsubmit is true, while preserving scheduled cron behavior.

go/traj/61432784-b795-4e4d-af0b-d60d97c7a0ae

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,13 +5,6 @@ on:
     - cron: "0 5 * * *" # Once a day at 5am.
   # Manual runs through Actions tab in the UI
   workflow_dispatch:
-    inputs:
-      any-postsubmit:
-        description: >-
-          Allow releasing from any successful postsubmit (including manual
-          runs), not just scheduled ones.
-        type: boolean
-        default: false
 
 permissions:
   actions: read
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -26,11 +33,16 @@ jobs:
         id: latest-green
         env:
           REPO: ${{ github.repository }}
+          ANY_POSTSUBMIT: ${{ inputs.any-postsubmit }}
         run: |
           set -euo pipefail
+          EVENT_FILTER="&event=schedule"
+          if [[ "${ANY_POSTSUBMIT}" == "true" || ("${{ github.event_name }}" == "workflow_dispatch" && "${ANY_POSTSUBMIT}" == "") ]] ; then
```
Tobias-Pe: [STYLE] Prefer passing `${{ github.event_name }}` via the `env` block (consistent with `REPO` and `ANY_POSTSUBMIT`) rather than interpolating workflow expressions directly inside the shell script. Additionally, ensure grouping parentheses in `[[ ... ]]` are surrounded by whitespace and format the clause as `]]; then` to follow shell style conventions (see http://go/shellstyle#control-flow).

🤖 **AI Reviewer** (Jetski) | Traces: style: http://go/traj/3044f789-f00e-4ebb-bfc0-18aabe8013c9

.github/workflows/release.yml
```diff
@@ -5,6 +5,13 @@ on:
     - cron: "0 5 * * *" # Once a day at 5am.
   # Manual runs through Actions tab in the UI
   workflow_dispatch:
+    inputs:
```
faximan: I am not sure why you need this input instead of just removing the filter? Is there a use case for why you wouldn't want to release from latest green?

Because it is always from main, I think you are guaranteed that latest green is also newest repo state, which is what would be released on the next scheduled run anyways.

Tobias-Pe: I didn't want to deviate too much from the current way, but yes would also work i think.

faximan (approved):

ensonic (approved):

PR #890, opened 2026-08-24
Support fast_push in run-install

koonpeng:
- Support `fast_push` in `run-install.sh`.
- Remove use of terraform in `fast_push`.

This is to allow projects that provision CRC directly via terraform module to update CRC charts using the deploy script.

### DIFF
--- a/src/bootstrap/cloud/run-install.sh
+++ b/src/bootstrap/cloud/run-install.sh
@@ -29,6 +29,7 @@ if [[ -z "${GCP_PROJECT_ID}" || "${GCP_PROJECT_ID}" == -* ]]; then
   echo "  --set-oauth     Enables and configures OAuth interactively."
   echo "  --delete        Deletes Cloud Robotics from the cloud project."
   echo "  --terraform     Apply only terraform changes."
+  echo "  --fast-push     Only deploy charts, skip terraform."
   echo "  (default)       Install the specified version."
   exit 1
 fi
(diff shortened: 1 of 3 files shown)


src/bootstrap/cloud/run-install.sh
```diff
@@ -63,11 +64,13 @@ if [[ $SHELLOPTS =~ xtrace ]] ; then
...
+if [[ "${COMMAND}" == "--set-config" ]]; then
   $BASH scripts/set-config.sh "${GCP_PROJECT_ID}" "$@"
-elif [[ "${COMMAND}" = "--set-oauth" ]]; then
+elif [[ "${COMMAND}" == "--fast-push" ]]; then
+  $BASH ./deploy.sh fast_push "${GCP_PROJECT_ID}" "$@"
+elif [[ "${COMMAND}" == "--set-oauth" ]]; then
   $BASH scripts/set-config.sh "${GCP_PROJECT_ID}" --edit-oauth "$@"
-elif [[ "${COMMAND}" = "--delete" ]]; then
+elif [[ "${COMMAND}" == "--delete" ]]; then
   $BASH ./deploy.sh delete "${GCP_PROJECT_ID}" "$@"
```
ensonic: shall we leave the `set-config` and `deploy` calls grouped. Or maybe switch to a case statement on the $COMMAND?

ensonic (approved):
Thanks. It's a pity that we can't keep accessing the terraform outputs.

src/bootstrap/cloud/run-install.sh
koonpeng: updated to use switch statement, and move `--fast-push` down with the deploy calls.

koonpeng:
> Thanks. It's a pity that we can't keep accessing the terraform outputs.

I think with tf managed config, we can export the ingress ip in `config.sh`. Then in `deploy.sh` we can try to read the config first, then fallback to terraform outputs.

ensonic (approved):

PR #900, opened 2026-08-28
Export service accounts

koonpeng:
Export service accounts so downstream can reference them without querying GCP or remote state.


ensonic (commented):
Good idea

ensonic (approved):

PR #902, opened 2026-08-31
Add a test covering the chunked transfer

ensonic:
Add unit and integration tests covering HTTP/1.1 chunked request bodies
(Transfer-Encoding: chunked) through the relay server:
- Unit test in server_test.go using io.Pipe and sync.WaitGroup to verify
  in-memory chunked request buffering and forwarding.
- End-to-end integration test in http_test.go using io.Pipe.
- Update Tested capabilities section in http-relay-server README.md.

### DIFF
--- a/src/go/tests/relay/http_test.go
+++ b/src/go/tests/relay/http_test.go
@@ -130,3 +130,67 @@ func TestHttpErrorPropagation(t *testing.T) {
 		})
 	}
 }
+
+func TestHttpChunkedRequestBody(t *testing.T) {
+	expectedRequestBody := []byte("chunk1-chunk2-chunk3")
+	expectedResponseBody := []byte("received: chunk1-chunk2-chunk3")
+
+	env := setupRelay(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


src/go/tests/relay/http_test.go
```diff
@@ -130,3 +130,69 @@ func TestHttpErrorPropagation(t *testing.T) {
...
+	pr, pw := io.Pipe()
+	go func() {
+		defer pw.Close()
+		chunks := [][]byte{[]byte("chunk1-"), []byte("chunk2-"), []byte("chunk3")}
+		for _, chunk := range chunks {
+			if _, err := pw.Write(chunk); err != nil {
+				t.Errorf("failed to write chunk to pipe: %v", err)
+				return
+			}
+			time.Sleep(10 * time.Millisecond)
```
csieber: Can we use synctest here to not have the sleep be an actual sleep? It might not work for all http flows but it could be worth a try.

ensonic: good point. Reworked the change to use a synctest (server_test), for the end to end test (http_test) this does not work since real network stack is used..

csieber (approved):

PR #903, opened 2026-08-31
http-relay-server: decouple request stream cleanup from broker mutex

ensonic:
In StopRelayRequest and ReapInactiveRequests, removeRequest was being called while holding the global broker lock (broker.m). Since removeRequest acquires per-request stream locks (requestStreamMutex and sendMutex) which may be held while a client is actively reading or sending a chunk, holding broker.m caused broker-wide lock congestion and stalled unrelated requests.

This change removes the request from the resp map under broker.m, releases the lock, and then performs channel closing and cleanup outside the lock.

TAG=agy
CONV=054fbbd3-c4cf-4e7b-9519-e9ab2a4413a7

### DIFF
--- a/src/go/cmd/http-relay-server/server/broker.go
+++ b/src/go/cmd/http-relay-server/server/broker.go
@@ -204,9 +204,14 @@ func (r *broker) RelayRequest(server string, request *pb.HttpRequest) (<-chan *p
 // with the relay id to not be recognized, resulting in the relay server returning an error.
 func (r *broker) StopRelayRequest(requestId string) {
 	r.m.Lock()
-	defer r.m.Unlock()
-	if pr, ok := r.resp[requestId]; ok {
-		r.removeRequest(requestId, pr)
+	pr, ok := r.resp[requestId]
+	if ok {
+		delete(r.resp, requestId)
+	}
+	r.m.Unlock()
+
+	if ok {
+		r.removeRequest(pr)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


src/go/cmd/http-relay-server/server/broker.go
```diff
@@ -348,17 +353,23 @@ func (r *broker) SendResponse(resp *pb.HttpResponse) error {
...
+
+	for _, pr := range toReap {
+		r.removeRequest(pr)
+	}
 }
 
-func (r *broker) removeRequest(id string, pr *pendingResponse) {
+func (r *broker) removeRequest(pr *pendingResponse) {
 	// Closing `pr.markReap` tells `SendResponse` and `PutRequestStream` to stop.
 	close(pr.markReap)
```
koonpeng, line 374: iirc the this should in theory signal to the broker to stop the request asap, which should cause this function to finish "quickly". But I don't see any issue with this PR, it removes the ambiguity and the edge case where the server is not able to end the request quickly.

koonpeng (approved):

src/go/cmd/http-relay-server/server/broker.go
ensonic: I don't understand the first sentence. Happy to chat on additional refactors on the signalling. For now I'd like improve the locking (will send 2 more PR later the week)

koonpeng: I'm just trying to recall why it was designed like this to understand if this change will cause any deadlocks.

The original design is that in `func (r *broker) removeRequest(pr *pendingResponse)`

```go
close(pr.markReap)
```

Should make `SendResponse` and `PutRequestStream` to stop asap. That should make `removeRequest` finish asap, minimizing the time `func (r *broker) ReapInactiveRequests(threshold time.Time)` holds the global mutex.

But that is an optimistic assumption, some edge cases may cause the global mutex to be locked longer than it needs to be.

ensonic: Thanks for the context and explanation, but I think the new code make it easier to see fro what access the lock is held.

PR #904, opened 2026-08-31
Add metric for legacy cr-syncer requests

drigz:
This will let us check it's safe to turn down the old codepath. This has
been tested with the new unit tests, I'm planning to double-check once
this is deployed to dev systems later this week.

I considered removing the tests before sending for review, but figured it's better to remove then when removing the old flag.


drigz:

ensonic (approved):

PR #910, opened 2026-09-01
Optimize broker lock handling in backend cleanup.

ensonic:
Collect inactive server names and remove them from r.req under r.m.Lock(), then release r.m before invoking Prometheus deletions and logging.


csieber (approved):
good idea!

PR #913, opened 2026-09-04
synk: prioritize RBAC resources in newGvknn sorting

ensonic:
Assign higher sort priorities to ClusterRole, Role, ClusterRoleBinding,
and RoleBinding in newGvknn so that RBAC rules are applied after
Namespaces, ServiceAccounts, and Secrets, and before remaining
workloads

TAG=agy
CONV=d1d5e845-2765-44b2-ab59-1fa15cb5878d

### DIFF
--- a/src/go/pkg/synk/synk.go
+++ b/src/go/pkg/synk/synk.go
@@ -324,8 +324,58 @@ func (s *Synk) applyAll(
 ) (applyResults, error) {
 	results := applyResults{}
 
-	crds, regulars := separateCRDsFromResources(resources)
+	groups := groupResources(resources, isCustomResourceDefinition, isRBACResource)
+	crds := groups[0]
 
+	if err := s.applyCRDs(ctx, rs, opts, results, crds); err != nil {
+		return results, err
+	}
+
+	for _, group := range groups[1:] {
+		s.applyResources(ctx, rs, opts, results, group)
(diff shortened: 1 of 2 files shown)


ensonic:
Is this overkill?

awhdesmond:
> Is this overkill?

Will it be cleaner if we specify the ordering in 

```go
// sort.go
func newGvknn(group, version, kind, namespace, name string) *gvknn {
```

drigz:
Afraid I won't have time to properly review this so removing myself as reviewer (feel free to re-add me if you can wait until next week or so). It seems reasonable in principle as a workaround for broken applications that don't retry these errors, I don't think there's any reason we'd want to install RBAC after other resources.

ensonic:
> > Is this overkill?
> 
> Will it be cleaner if we specify the ordering in
> 
> ```go
> // sort.go
> func newGvknn(group, version, kind, namespace, name string) *gvknn {
> ```

Totally, this is much better. PTAL.

awhdesmond (approved):
LGTM.
remaining small nit: PR desc can be updated to reflect the new changes.

ensonic:
> LGTM. remaining small nit: PR desc can be updated to reflect the new changes.

Done. The commit was updated. Copied into the github PR msg.

PR #914, opened 2026-09-04
Onboard renovate bot to help with bazel maintenance.

ensonic:
It will run via github action (see https://github.com/renovatebot/github-action).

### DIFF
--- a/.github/ci/renovate-config.js
+++ b/.github/ci/renovate-config.js
@@ -1,0 +1,120 @@
+/**
+ * @see https://docs.renovatebot.com/configuration-options/
+ * @type {import("renovate/dist/config/types").RenovateConfig}
+ */
+export default {
+  extends: ['config:best-practices'],
+  hostRules: [
+    ...(process.env.RENOVATE_GCR_TOKEN
+      ? [
+          {
+            hostType: 'docker',
+            matchHost: 'gcr.io',
+            authType: 'Basic',
+            token: process.env.RENOVATE_GCR_TOKEN,
+          },
+        ]
+      : []),
+  ],
+  repositories: ['googlecloudrobotics/core'],
+  enabledManagers: [
+    'bazel-module',
+    'bazelisk',
+    'custom.regex',
+    'dockerfile',
+    'github-actions',
+  ],
+  customManagers: [
+    {
+      customType: 'regex',
+      managerFilePatterns: [
+        '.github/workflows/renovate.yml',
+        'MODULE.bazel',
+        'non_module_deps.bzl',
+        '.bazelversion',
+        '.github/ci/Dockerfile.integration-test-image',
+      ],
+      matchStrings: [
+        // Match annotated versions with commit SHA digest and release tag comment, e.g.:
+        //   # renovate: datasource=github-releases dep_name=golang/tools
+        //   - golang.org/x/tools/cmd/goimports@24a8e95f9d7ae2696f66314da5e50c0d98ccaa90 # v0.43.0
+        `(?:#|\\/\\/) renovate: datasource=(?<datasource>\\S*) dep_name=(?<depName>\\S*)(?: versioning=(?<versioning>\\S*))?\\s*[\\s-]*\\S*?(?::\\s?|\\s?=\\s?|@)(?:"|')?(?<currentDigest>[0-9a-f]{40})(?:"|')?\\s+#\\s*(?<currentValue>\\S+)`,
+        // Match annotated versions.
+        //
+        // Discovers marked versions by looking for a `renovate:` comment. This
+        // comment looks like this:
+        //
+        //    # renovate: datasource=github-releases dep_name=org/repo versioning=loose
+        //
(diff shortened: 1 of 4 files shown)


.github/workflows/renovate.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          sudo mkdir -p /tmp/renovate
+          sudo chown -R 12021:0 /tmp/renovate
+
+      - name: Generate app token
+        id: renovate-token
+        if: ${{ env.RENOVATE_APP_ID != '' }}
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.RENOVATE_APP_ID || secrets.INTRINSIC_RENOVATE_APP_ID }}
```
faximan, line 73: I only see `RENOVATE_APP_ID ` and `RENOVATE_APP_PRIVATE_KEY ` as secrets in https://github.com/googlecloudrobotics/core/settings/secrets/actions. Should we remove all other versions here? E.g. the mentions of INTRINSIC.

faximan, line 73: nit: GitHub wants us to use `client-id` instead of `app-id` these days. See e.g. migrations in https://github.com/intrinsic-ai/insrc/pull/52930/changes.

You can find the client ID on https://github.com/organizations/googlecloudrobotics/settings/apps/googlecloudrobotics-renovate.

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,106 @@
...
+      # BUG https://github.com/renovatebot/github-action/issues/646
+      # user: ubuntu(12021), group: root(0)
+      - name: Fix cache permissions
+        run: |
+          sudo mkdir -p /tmp/renovate
+          sudo chown -R 12021:0 /tmp/renovate
+
+      - name: Generate app token
+        id: renovate-token
+        if: ${{ env.RENOVATE_APP_ID != '' }}
```
faximan, line 70: I am not sure this works. I think `if` conditionals are evaluated before the env is loaded, meaning this will never execute.

Do you want something like

```
   env:
          RENOVATE_APP_ID: ${{ secrets.RENOVATE_APP_ID }}
```

.github/workflows/renovate.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          echo "RENOVATE_GITHUB_ACTOR=${{ github.actor }}" >> "$GITHUB_ENV"
+          JSON_KEY_BASE64="$(echo -n "oauth2accesstoken:${{ steps.gcp-auth.outputs.access_token }}" | base64 -w 0)"
+          echo "RENOVATE_GCR_TOKEN=${JSON_KEY_BASE64}" >> "$GITHUB_ENV"
+
+      - name: Run Renovate to check for updated dependencies
+        uses: renovatebot/github-action@39b914146caeff8cd512e61c8992f1d5913af85c # v46.2.5
+        with:
+          configurationFile: .github/ci/renovate-config.js
+          token: "${{ steps.renovate-token.outputs.token || secrets.RENOVATE_TOKEN || github.token }}"
```
faximan, line 100: For `github.token` I think you need `contents: write` in `permissions`.

Although maybe relying on `steps.renovate-token.outputs.token` is enough.

.github/ci/renovate-config.js
```diff
@@ -0,0 +1,120 @@
...
+  ],
+  packageRules: [
+    {
+      matchDatasources: ['golang-version'],
+      matchDepNames: ['go'],
+      groupName: 'golang',
+    },
+    {
+      groupName: 'renovate',
+      matchDepNames: ['renovate', 'ghcr.io/renovatebot/renovate'],
```
faximan, line 90: Nit: I think this should also include `renovatebot/github-action` so that the action is updated together. It probably doesn't matter much for CRC because I don't think you are using the other two versions, but maybe we should to make this change in insrc too.

faximan (commented):
Thanks Stefan!

.github/workflows/renovate.yml
ensonic: Done

.github/workflows/renovate.yml
ensonic: Removed the if, the env part is already there

.github/workflows/renovate.yml
ensonic: I'd rather keep this aligned with the other workflows for a start. SGTY?

.github/ci/renovate-config.js
ensonic: TBH, I don't know what change your're asking for. Can we try to onboard it first and then make the change in both places if it makes sense?

.github/workflows/renovate.yml
faximan: Sure we can, I don't think it will cause any problems even if it is broken.

faximan (approved):
Thanks @ensonic ! Let's merge this and iterate as needed.

PR #946, opened 2026-09-14
Readd MODULE.bazel.lock

ensonic:
We lost this in one of the prev changes, but It need to be part of the gitrepo.


faximan (approved):
>  It need to be part of the gitrepo.

I don't think it needs to be (our internal repos don't have it), but it might be a good idea to check it in if users are prepared to handle merge conflicts (https://bazel.build/external/lockfile#merge-conflicts).

ensonic:
@faximan FYI: https://bazel.build/external/lockfile#best-practices

PR #956, opened 2026-09-15
Use domain-qualified finalizer name for ChartAssignment

ensonic:
Update the ChartAssignment finalizer from "helm.apps.cloudrobotics.com" to "helm.apps.cloudrobotics.com/chartassignment" to satisfy Kubernetes API validation requiring custom finalizer names to be domain-qualified with a path (/).

Automatically migrate existing ChartAssignment resources away from the deprecated finalizer name during reconciliation, and ensure both finalizer names are cleaned up when deleting a ChartAssignment.

See Kubernetes documentation on custom resource finalizers and naming:
- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers
- https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/

TAG=agy
CONV=76ef7db7-aae5-4eaf-8ed8-4b1e51c8a629


src/go/pkg/controller/chartassignment/controller.go
```diff
@@ -328,9 +330,15 @@ func (r *Reconciler) reconcile(ctx context.Context, as *apps.ChartAssignment) (r
 	}
 	// Ensure a finalizer on the ChartAssignment so we don't get deleted before
 	// we've properly deleted the associated Synk ResourceSet.
-	if !stringsContain(as.Finalizers, finalizer) {
-		as.Finalizers = append(as.Finalizers, finalizer)
+	if !stringsContain(as.Finalizers, finalizer) || stringsContain(as.Finalizers, deprecatedFinalizer) {
```
therjak: this code looks enough like a bug (missing ! on the second stringsContain) that I would add a comment to make clear this is intended.

ensonic: Agree. Splitted the logic to make it more readable. PTAL.

therjak (approved):

PR #958, opened 2026-09-15
Update rules_oci to v2.3.0 and pass RUNFILES_DIR in push wrappers

ensonic:
Starting with rules_oci v2.2.7+, oci_push uses Bazel's runfiles.bash v3 and rlocation instead of relative paths from .runfiles/_main. Propagate RUNFILES_DIR (in addition to legacy PYTHON_RUNFILES) across the push wrapper rules (cache_gcr_credentials, run_parallel, and push_all) so that sub-binaries invoked inside runfiles can locate runfiles.bash.

Also remove the dummy repository placeholders on oci_push targets now that rules_oci v2.3.0 supports omitting the repository attribute at build time.

TAG=agy
CONV=0b7ae56c-4b44-4f5c-ba68-e6c26f1c036b


ensonic:
This change is a close of https://github.com/googlecloudrobotics/core/pull/935, but with additional fixes.

csieber (approved):

PR #965, opened 2026-09-16
Split k8s_integration_test and add synctest/NewTestServer unit tests

ensonic:
Extract cluster health and AppRollout polling logic into k8s_check.go using wait.PollUntilContextTimeout instead of manual time.Sleep loops.

Add in-memory tests in k8s_check_test.go using httptest.NewTestServer and testing/synctest under a non-external //src/go/tests:k8s_check_test target so they run locally and in presubmit in ~0.1s, while keeping the live-cluster tests in //src/go/tests:k8s_integration_test tagged as external.

TAG=agy
CONV=ba8f0331-fd44-462f-9ef9-e0eba806dada


src/go/tests/k8s_check.go
```diff
@@ -0,0 +1,197 @@
+// Copyright 2019 The Cloud Robotics Authors
```
therjak: please use 2026

src/go/tests/k8s_check.go
```diff
@@ -0,0 +1,197 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	AppInitializationTimeout = 7 * time.Minute
```
therjak: do these variables need to be exported? i only see use within this package. same with the functions that previously were not exported but are now

src/go/tests/k8s_check.go
```diff
@@ -0,0 +1,197 @@
...
+		slog.Info("...done.", slog.Int("PodCount", len(pods.Items)))
+
+		if len(pods.Items) == 0 {
+			return false, fmt.Errorf("could not find any pods in cluster")
+		}
+
+		numNonRunningPods = 0
+		failingContainers = 0
+		for _, pod := range pods.Items {
+			slog.Info("Pod state", slog.String("Name", pod.Name), slog.String("Phase", string(pod.Status.Phase)))
```
therjak: as the function has a context, you should probably use slog.InfoContext and slog.WarnContext

src/go/tests/k8s_check.go
ensonic: done

src/go/tests/k8s_check.go
ensonic: done.

therjak (approved):

PR #972, opened 2026-09-17
Add default scrape class with cluster relabelling

awhdesmond:
1. Add default scrape class with cluster relabelling to add local cluster label for cloud scrape jobs.
2. Will not affect robots federation as relabelling gated via regex looking for empty labels.


src/app_charts/prometheus/prometheus-cloud.values.yaml
```diff
@@ -59,6 +59,15 @@ prometheus:
...
         operator: "DoesNotExist"
 
+    scrapeClasses:
+    - name: default
+      default: true
+      metricRelabelings:
+      - sourceLabels: [cluster]
+        regex: '^$'
+        targetLabel: cluster
+        replacement: "cloud"
```
awhdesmond: replacement can be changed. let me know which you all prefer.

ensonic: maybe cloud-robotics to tell apart from the ml-compute cluster
```suggestion
        replacement: "cloud-robotics"
```

ensonic (approved):

src/app_charts/prometheus/prometheus-cloud.values.yaml
awhdesmond: done

PR #978, opened 2026-09-18
store pub-key configmaps in default ns with Robot Owner References

ensonic:
Migrate token-vendor public key ConfigMaps from app-token-vendor to the default namespace at startup and configure token-vendor to use the default ns. When storing or migrating a public key ConfigMap, check if a matching Robot CR (stripping the 'robot-' prefix) exists in the default namespace and attach a Kubernetes OwnerReference so the ConfigMap is automatically garbage-collected when the Robot CR is deleted.

TAG=agy
CONV=cdf89e3c-af60-4993-a6a4-ef1dc97f1eaa


therjak (approved):

src/go/cmd/token-vendor/repository/k8s/k8s.go
```diff
@@ -84,15 +138,155 @@ const (
...
+	robotPrefix    = "robot-"
 )
 
+func isDeviceConfigMap(cm *corev1.ConfigMap) bool {
+	if cm.Labels[managedByLabel] == managedByValue {
+		return true
+	}
+	_, hasPubKey := cm.Data[pubKey]
+	return hasPubKey
+}
```
ensonic: Either now or in a folowup, I can add the filtering directly to the informer.

therjak (approved):

PR #982, opened 2026-09-18
synk: add benchmarks + optimize sorting, discovery, and concurrent apply

ensonic:
- Add in-process benchmark suite in src/go/pkg/synk/benchmark_test.go
  covering sortResources, Synk.Apply create/update, API/discovery call
  counts, and simulated network latency.
- Avoid heap allocations and fmt.Sprintf in newGvknn and less.
- Apply CRDs and priority tiers concurrently in applyAll (up to 16 workers)
  while preserving strict ordering across priority tiers.
- Only invalidate the discovery cache and reset RESTMapper when a CRD is
  not already present in discovery.
- Cache strategicpatch.PatchMetaFromStruct by GroupVersionKind in applyOne.
    
Performance Improvements:
- BenchmarkSortResources (200 resources): 1.77ms -> 0.41ms (4.3x faster),
  allocations reduced from 21,611 to 2 allocs/op (521 KB -> 3.6 KB/op).
- BenchmarkApply_Update (no-op re-apply, 100 resources + 6 CRDs):
  25.1ms -> 13.7ms (1.8x faster), discovery calls reduced from 7 to 1.
- BenchmarkApply_SimulatedLatency (1ms RTT, 100 resources + 6 CRDs):
  125.8ms -> 26.0ms (4.8x faster).
- End-to-end `./deploy.sh fast_push` on a live GKE cluster:
  total runtime dropped from 1m25.3s to 36.3s (2.35x faster); the synk
  chart sync phase dropped from ~56s to ~10s (~5.5x faster).

TAG=agy
CONV=e5127a19-1491-4883-8102-b3a1a652ffea


src/go/pkg/synk/synk.go
```diff
@@ -323,70 +331,98 @@ func (s *Synk) applyAll(
...
-		// CRDs must never be replaced as deleting them will delete
-		// all its current instances. Update conflicts must be resolved manually.
-		action, err := s.applyOne(ctx, crd, rs)
-		if err != nil {
-			opts.errorf(crd, action, "failed to apply: %s", err)
-		} else {
-			opts.logf(crd, action, "applied successfully")
+	if len(crds) > 0 {
+		// Insert CRDs and wait for them to become available.
+		sem := make(chan struct{}, workers)
```
therjak: i guess 'sem' is abbreviated for semaphore? I would avoid abbreviations where the name carries information on what it actually is. I had to deduct the name from how it is used but I think it should be the other way around.

src/go/pkg/synk/synk.go
```diff
@@ -323,70 +331,98 @@ func (s *Synk) applyAll(
...
-	// an upper bound just in case of flapping errors.
-	prevFailures := 0
+	for _, tier := range groupByPriority(regulars) {
+		// Try applying until the errors stay the same between iterations. Put in
+		// an upper bound just in case of flapping errors.
+		prevFailures := 0
 
-	for i := 0; i < 10; i++ {
-		curFailures := 0
+		for i := 0; i < 10; i++ {
```
therjak, line 391: i guess the new way (go 1.22+) is to write 'for i := range 10'

src/go/pkg/synk/synk.go
```diff
@@ -323,70 +331,98 @@ func (s *Synk) applyAll(
...
+			return results, fmt.Errorf("wait for CRDs: %w", err)
+		}
 	}
-	// Reset all discovery and mapping once again.
-	s.resetMapper()
 
-	// Try applying until the errors stay the same between iterations. Put in
-	// an upper bound just in case of flapping errors.
-	prevFailures := 0
+	for _, tier := range groupByPriority(regulars) {
```
therjak, line 386: is regulars already sorted by priority? It does look like groupByPriority needs them sorted as it would create a new tier for every priority change in the slice. so has groupByPriority and iterating over the tiers even an effect? to me it looks like the order of execution did not change

ensonic: Yes, regulars is already sorted by priority (Apply() calls sortResources(resources) in initialize() before applyAll).

src/go/pkg/synk/synk.go
ensonic: Renamed to activeWorkers.

src/go/pkg/synk/synk.go
therjak: i still do not see why groupByPriority exists. if I got this right is just changes
[a b c d e] into something like [a b] [c d] [e] and executes it again as [a b c d e].

ensonic: The wg.Wait() ensures that the group is finished before processing to the next group. If we would no do this, resources from group N could still be processed while we already start with resources of group N+1

therjak (approved):

PR #985, opened 2026-09-21
add envoy gateway tf input

koonpeng:


src/bootstrap/cloud/terraform/input.tf
```diff
@@ -117,6 +117,12 @@ variable "enable_nginx_shield" {
   default = false
 }
 
+variable "use_envoy_gateway" {
+  description = "Use envoy gateway in place of nginx ingress"
```
awhdesmond, line 121: should we just use `use_gateway`?

koonpeng: done

awhdesmond (approved):

